PR #19, opened 2022-11-03
Add function to /bin to build clouds docs with bccutil / ccutil

ariordan-redhat:
This repo does not contain asciidoc artefacts such as `master.adoc`.
These artefacts are in the production repo: they are required for the build process for the customer portal.

This PR adds a function that fetches the artefacts from the production repo, builds the doc with ccutil / bccutil, and deletes the artefacts.

### DIFF
--- a/bin/aap-docs-bccutil.sh
+++ b/bin/aap-docs-bccutil.sh
@@ -1,0 +1,42 @@
+# Since master.adoc and docinfo.xml are not in the aap-docs repo, you can't use bccutil / ccutil to build aap-docs.
+# This function fetches files from Gitlab, runs bccutil, and deletes the added files.
+# 
+# Add this function to your ~/.zshrc or ~/.bashrc file, then run source ~/.bashrc or source ~/.zshrc
+#
+# Prerequisites:
+# * You must be connected to the VPN to run this function, as it fetches files from Gitlab.
+#   Error message when you run function when not connected to the VPN:
+#     ssh: Could not resolve hostname gitlab.cee.redhat.com: nodename nor servname provided, or not known
+# * You must have set up a git remote called "gitlab" for the Gitlab production repo. 
+#   To check your git remotes, run the following command:
+#     git remote -v
+#   Error message when you run function without gitlab remote set up:
+#     fatal: 'test' does not appear to be a git repository
+#     fatal: Could not read from remote repository.
+#
+# Procedure:
+# Run this function from a title directory, for example aap-on-aws
+# To run this function, type aap-docs-build <title-directory>, for example:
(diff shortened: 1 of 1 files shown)


bin/aap-docs-bccutil.sh
```diff
@@ -0,0 +1,26 @@
+# Since master.adoc and docinfo.xml are not in the aap-docs repo, you can't use bccutil / ccutil to build aap-docs.
+# This function fetches files from Gitlab, runs bccutil, and deletes the added files.
+# 
+# Add this function to your ~/.zshrc or ~/.bashrc file, then run source ~/.bashrc or source ~/.zshrc
+#
+# Prerequisites:
+# You must be connected to the VPN to run this function, as it fetches files from Gitlab.
+# You must have set up a git remote called "gitlab" for the Gitlab production repo.
```
ianf77, line 8: I'd add a 'what if you aren't' bit, and how to check whether you are.

bin/aap-docs-bccutil.sh
```diff
@@ -0,0 +1,26 @@
...
+# aap-docs-build aap-on-aws
+# aap-docs-build aap-on-gcp
+#
+    function aap-docs-build () {
+    git fetch gitlab aap-clouds-latest
+    printf "*** Building $1 docs ***\n"
+    git show gitlab/aap-clouds-latest:$1/master.adoc > master.adoc
+    git show gitlab/aap-clouds-latest:$1/docinfo.xml > docinfo.xml
+    git show gitlab/aap-clouds-latest:$1/title-attributes.adoc > title-attributes.adoc
+    # In the following line, substitute the command that you use to build the docs with ccutil / bccutil. The example here is for a mac. Delete the hashtag before the command.
```
ianf77, line 23: At what point? If it's the runCCutil bit, maybe have it as <runCCutil>?

ariordan-redhat: @ianf77 What command do you use to run ccutil?
I'm on a mac, so I use "leviathin"

ianf77: At the level of the document, so where I'd find master.adoc I'd run bccutil compile --lang en-US --format html-single --open, though I now have an alias as that's a bit long-winded, so now I just use docb.

ariordan-redhat: Thanks: I'll add in your full command as an example for RHEL.

ariordan-redhat: @ianf77 
I added example commands for mac and RHEL.
Does this section make sense now?

ariordan-redhat:
Function not needed.

PR #32, opened 2022-11-29
extension pointers and name limit

robinbobbitt:


robinbobbitt:
Changes were included in #34

PR #42, opened 2022-12-07
Update firewall rule requirements to globally access quay.io

scottharwell:
Update reference to `quay.io` in firewall rules to enable all subdomains that may be used.


ariordan-redhat (approved):
Looks good to me

PR #43, opened 2022-12-07
Re-adding quay.io for Azure firewall rule requirement

benjdudas:
Adding `quay.io` back with the addition of quay cdn match that was added with https://github.com/ansible/aap-docs/pull/42


ariordan-redhat (approved):
Looks good to me

PR #54, opened 2023-01-25
Added disclaimer

ianf77:
Add legal disclaimer for content of external sites.


ianf77:
I wasn't sure where it should go. I suppose we should be consistent through any document that contains this. Maybe in master.adoc? But that would be a change to the gitlab file here.

ariordan-redhat (approved):
LGTM

PR #61, opened 2023-02-15
Add domains to outbound routing list

ariordan-redhat:
Expand the domain list in the _Outbound routing through virtual appliances_ section of the AAP on Azure docs.


ianf77 (commented):
LGTM, though it mentions quay, and I was told by Richard Griffiths that quay was only used upstream and should never appear in downstream docs. Might be worth checking?

ianf77 (approved):
LGTM

PR #63, opened 2023-02-16
Restructure new aws content

ianf77:


stories/topics/proc-aap-aws-modify-cidr-ranges.adoc
```diff
@@ -1,42 +1,57 @@
 [id="proc-aap-aws-modify-cidr-ranges"]
 
-= Modify Deployment CIDR Range
+= Modifying the VPC CIDR range after deployment
```
zkayyali812, line 3: Instead of `VPC` maybe we can use `Application`, since we will modify the CIDR range in not only the VPC, but also subnets.

zkayyali812: Also, the title here is `Modifying the VPC CIDR range after deployment`, but updating CIDR after deployment is not supported, and requires the stack to be deleted and redeployed to update the CIDR.

Perhaps we can call this `Modifying the CIDR range before deployment` and draw attention to the note that says it cant be modified after.

zkayyali812 (changes requested):
A few small comments! Thanks the refactor here 😄

zkayyali812 (approved):
/lgtm!

stories/assembly-aap-aws-intro.adoc
```diff
@@ -18,5 +18,9 @@ The following {PlatformName} components are available on {AAPonAWS}:
 * {ExecEnvNameStart}
 * Ansible content tools, including access to {InsightsName}
 
+[NOTE]
+=====
+{AutomationMeshStart} is not available on {AAPonAWS} at this time and is planned for a later release.
```
ariordan-redhat, line 23: @ianf77 
Would it be better not to mention that Mesh is planned for a later release?

ianf77: I can do.

ariordan-redhat (approved):
LGTM

PR #69, opened 2023-03-02
Corrected issues in AWS chapters

ianf77:
Spring cleaning

### DIFF
--- a/stories/assembly-aap-aws-backup-and-recovery.adoc
+++ b/stories/assembly-aap-aws-backup-and-recovery.adoc
@@ -5,4 +5,12 @@ ifdef::context[:parent-context: {context}]
 
 :context: aap-aws-backup-recovery
 
+To backup and restore your {PlatformNameShort} deployment, it is vital to ensure automatic backups have been set for EFS and that these backups are accessible for restoration.
+
+When you create an {PlatformNameShort} deployment, automatic Amazon Elastic File System (EFS) backups are set by default, however it is prudent to check if those backups can be restored before a disaster scenario.
+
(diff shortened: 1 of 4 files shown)


stories/assembly-aap-aws-backup-and-recovery.adoc
```diff
@@ -5,4 +5,12 @@ ifdef::context[:parent-context: {context}]
 
 :context: aap-aws-backup-recovery
 
+To backup and restore your {PlatformNameShort} deployment, it is vital to ensure automatic backups have been set for EFS and that these backups are accessible for restoration.
```
ariordan-redhat, line 8: Should you define EFS here instead of in the 2nd paragraph?

stories/assembly-aap-aws-backup-and-recovery.adoc
```diff
@@ -5,4 +5,12 @@ ifdef::context[:parent-context: {context}]
...
+
+When you create an {PlatformNameShort} deployment, automatic Amazon Elastic File System (EFS) backups are set by default, however it is prudent to check if those backups can be restored before a disaster scenario.
```
ariordan-redhat, line 10:
```suggestion
When you create an {PlatformNameShort} deployment, automatic Amazon Elastic File System (EFS) backups are set by default. However it is prudent to check that those backups can be restored before a disaster scenario.
```

stories/assembly-aap-aws-backup-and-recovery.adoc
ianf77: it's actually defined at the start  of the document as well.

ariordan-redhat (approved):
LGTM

PR #103, opened 2023-05-18
AWS Change aap-aws to gcp within all AWS modules and assemblies

ariordan-redhat:
- Change every instance of `aap-aws` in all modules and assemblies to `aws`.
- Rename AWS assemblies, images, and modules to eliminate `aap-`
- Update `stories.adoc` to use the new filenames


ianf77 (approved):
LGTM

Itt's shown up one or two non-related things that need fixing as well! Doubly useful!

PR #108, opened 2023-05-25
Azure - document disaster recovery

ariordan-redhat:
Change Azure storage shape when disaster recovery is enabled


stories/topics/con-azure-disaster-recovery.adoc
```diff
@@ -0,0 +1,11 @@
...
+
+= Disaster Recovery
+
+When you deploy {AAPonAzureNameShort}, you must enable or disable Disaster Recovery in the `Business Continuity` tab of the form.
+There is no default setting for Disaster Recovery.
+
+The Disaster Recovery feature incurs additional Azure infrastructure costs.
+See xref:con-azure-infrastructure-usage_azure-intro[Azure infrastructure usage] for details of the Service Shape of the Storage account.
+
+
```
aldato, line 11: We should add that once the AAP is deployed, customers who'd like to enable DI should contact customer support in order to do so.

aldato: Shall we add here a link to the KCS which explains Disaster Recovery in detail? 

https://access.redhat.com/articles/7010302

stories/topics/con-azure-infrastructure-usage.adoc
```diff
@@ -20,7 +20,10 @@ Azure Database for PostgreSQL:: An Azure database service used for {PlatformName
 **Service Shape:** 1 TB
 Storage account:: The Azure service is used for file and block storage for local storage of projects, containers, etc.
```
EMcWhinn, line 21: As per SG, I would avoid using "etc.", instead:

"The Azure service is used for file and block storage such as local storage of projects and containers."

stories/topics/con-azure-disaster-recovery.adoc
```diff
@@ -0,0 +1,17 @@
+[id="con-azure-disaster-recovery_{context}"]
+
+= Disaster Recovery
+
+When you deploy {AAPonAzureNameShort}, you must enable or disable Disaster Recovery in the `Business Continuity` tab of the form.
+There is no default setting for Disaster Recovery.
+
+If you want to enable Disaster Recovery on an existing instance of {AAPonAzureNameShort}, contact Red Hat customer support.
```
EMcWhinn, line 8: Just wondering here if we should hyperlink "Red Hat customer support" directly to support, or is this fairly common knowledge?

EMcWhinn:
Just added 2 very small comments, otherwise LGTM.

stories/topics/con-azure-disaster-recovery.adoc
ariordan-redhat: I thought it better not to link, as different customers might have different contact points with Red Hat.

dcdacosta (approved):
LGTM!

PR #110, opened 2023-05-30
Removing podidentity install requirement for AOC on Azure

benjdudas:
Since moving to workload identity in version 1.0.28 we no longer require customer to enable the pod identity feature prior to deployment. 

JIRA: https://issues.redhat.com/browse/AAP-12595


ariordan-redhat (approved):
LGTM
No xrefs to the deleted module; Docs build OK.

benjdudas:
> @benjdudas Do you want this published on the customer portal immediately?

Thanks @ariordan-redhat , if you could that would be great. We pushed out the new release yesterday so as of today this is no longer a install requirement.

PR #126, opened 2023-06-09
change asciidoc tag so that $ is not included when copy/paste used

chrisahl:
This is a test of the proper [source tag to use to get the copy/paste to work the way we want.   I can't truly test this unless I get an official generated draft image.

If this works, we will have a bunch of places we need to change (and some we need to keep as is)


chrisahl:
Going to close this out since we now know the answer is to use the following before the block of code:

```
[literal,subs="+quotes,attributes",options="nowrap",role=white-space-pre]
```

PR #128, opened 2023-06-09
Archive unused AWS and GCP topics

ariordan-redhat:
Archive unused AWS and GCP topics

Docs build OK locally after deletion of files.


ianf77 (approved):
LGTM

PR #132, opened 2023-06-12
2.2.1 Remove source,bash line preceding codeblocks

ariordan-redhat:
Remove `[source,bash]` line preceding codeblocks

Removing the [source,bash] asciidoc tag so that the `$` prompt is not included when the copy/paste icon is used in the published docs in the customer portal

Backports #131 to aap-clouds-2.2.1


ianf77 (approved):
LGTM

PR #136, opened 2023-06-14
2.4 Only: Add extra-playbooks documentation

itdove:
https://issues.redhat.com/browse/AAP-10828


zkayyali812 (commented):
Looks good!
We should not merge this into aap-clouds-latest now, but we should merge it into our feature branch once we have that created.
Let us leave in draft for now until this feature branch is ready.

After our 23 changes are in, we should tag doc team to review this doc to ensure it is correct.
We can also ask QE now to begin validating this content

PR #137, opened 2023-06-14
AWS Fix assembly ID

ariordan-redhat:
Fix the `id` for `stories/assembly-aws-networking-access.adoc` so that current external links referencing this chapter do not break.
No internal `xref` instances reference this assembly.
The path to the affected assembly is updated in a local build.


ianf77 (approved):
LGTM

PR #146, opened 2023-06-16
Delete source,bash from line preceding codeblocks

ariordan-redhat:
Remove `[source,bash]` line preceding codeblocks

Removing the [source,bash] asciidoc tag so that the `$` prompt is not included when the copy/paste icon is used in the published docs in the customer portal


chrisahl:
LGTM.  Hopefully dev team will learn to stop using [source,bash] soon!  Without having [source,bash] to copy/paste it should be easier to accomplish!  Need to retrain our brain...

ianf77 (approved):
LGTM

PR #156, opened 2023-06-21
update existing vpc

leena-jawale:


stories/topics/proc-gcp-deploy-with-existing-vpc.adoc
```diff
@@ -3,6 +3,10 @@
 = Deploying an application with an existing VPC
 
 The following procedure uses an existing VPC network to deploy an application.
+[NOTE]
+====
+The existing proxy subnet should be of type *Regional Managed Proxy* which is the reserved proxy-only subnet for load balancing.
```
ianf77, line 8: must be of type * Regional Managed Proxy *,

ianf77 (commented):
One small change.

PR #162, opened 2023-06-23
Fix images symlink in titles directory

ariordan-redhat:
Fix symlink in the titles directory to point to the correct directory


ianf77 (approved):
LGTM

PR #172, opened 2023-07-06
Added new architecture images

ianf77:
Added images for AWS full and existing deployments, and GCP full and existing deployments


ariordan-redhat (approved):
LGTM

PR #186, opened 2023-07-12
add and remove tags tech note

prat98:


zkayyali812:
@prat98 I think the tech note around backup and restore here can be removed as well.
Please confirm and remove in this PR or another PR if so.

prat98:
can confirm that is not an issue any more so it can be removed from the tech note section

chrisahl:
Where are these showing in the Doc?  I don't see a Tech Notes section

prat98:
It's under the technical notes section of the doc section 12. I see it when I generate the doc locally

chrisahl:
Thanks. I now see it way down in AWS...but not in GCP yet.

chrisahl (approved):

stories/topics/tech-note-aws-add-remove-tags.adoc
```diff
@@ -0,0 +1,7 @@
+[id="tech-note-aws-add-remove-tags"]
+
+= Add and Remove Tags
+
+{AAPonAWS} supports stack-level tagging during stack deployment time. A stack-level tag is applied to the stack itself and all tag supported components that make up the stack.
```
ianf77, line 5: Perhaps you might say what a tag is?

ianf77 (commented):
one comment

stories/topics/tech-note-aws-add-remove-tags.adoc
chrisahl: I think AWS users would be familiar with tags, but if you do need to add some text about it, here is one example AWS has:
image

ianf77: Tags are custom fields that you assign to an AWS resource to help organize and identify your instances. {AAPonAWS} supports stack-level tagging during stack deployment time. A stack-level tag, when applied to the stack itself, is also applied to all tag supported components that make up the stack.

stories/topics/tech-note-aws-add-remove-tags.adoc
```diff
@@ -0,0 +1,7 @@
...
+
+Using the add or remove tags playbooks will only update the tags for the individual components of stack and not the tag applied to the stack itself. The stack-level tags will never be modified after deploying {AAPonAWS}.
```
ianf77, line 7: playbooks only update the tags for...
Stack-level tags are never modified after deployment.

ianf77 (commented):
Some comments

stories/assembly-aws-tech-notes.adoc
```diff
@@ -8,3 +8,4 @@ ifdef::context[:parent-context: {context}]
 {AAPonAWS} is a self-managed deployment. The following are technical notes regarding the {AAPonAWS} deployment.
 
 include::topics/tech-note-aws-add-remove-tags.adoc[leveloffset=+1]
+include::topics/tech-note-ansible-on-clouds-ops-container.adoc[leveloffset=+1]
```
chrisahl, line 11: Do you need aws in file name:  tech-note-aws-ansible-on-clouds-ops-container.adoc to follow convention?

zkayyali812: Let us add the `aws` in there to maintain our naming convention like @chrisahl notes @prat98

ianf77 (approved):
LGTM

PR #190, opened 2023-07-13
no longer need this because we are going to hide the playbooks so the…

chrisahl:
… user cannot see them

Based on the fact that we are going to change the ops command generator code to no longer show the gcp external load balancer playbooks, we need to get rid of this doc text

See https://github.com/ansible/aap-aoc-collections/pull/326


zkayyali812 (approved):

ianf77 (approved):
LGTM

PR #191, opened 2023-07-13
remove gcp tech note

leena-jawale:


zkayyali812 (approved):
nice!
It is good to comment this out, so we can more quickly add tech notes when necessary.

Thanks Leena!

ianf77 (approved):
LGTM

PR #210, opened 2023-07-18
AWS Add Remove Extension Node Permissions and Tech Note

prat98:


stories/topics/tech-note-aws-add-remove-extension-nodes.adoc
```diff
@@ -0,0 +1,5 @@
+[id="tech-note-aws-add-remove-extension-nodes"]
+
+= Add and Remove Extension Nodes
+
+On rare occasions, the Anisble on Clouds Ops Container remove extension node playbook may fail with the message `Waited too long for old instances to terminate`. Simply running the playbook again should resolve the issue and allow all extension nodes attached to a {AAPonAWS} deployment to be deleted.
```
ianf77, line 5: Do you mean aws_remove_extension_nodes.yaml?

can fail with

Run the playbook again to  resolve the issue and enable you to delete all extension nodes attached to an {AAPonAWS} deployment.

ianf77 (commented):
a few comments

stories/topics/tech-note-aws-add-remove-extension-nodes.adoc
prat98: So I put in `Anisble on Clouds Ops Container remove extension node playbook` instead of `aws_remove_extension_nodes.yaml` because the customer interfaces with the Ops container as opposed to the playbooks directly. I know it's more of a mouthful but this would be something customer would be more familiar with.

ianf77 (approved):
LGTM

PR #225, opened 2023-07-21
Fix xrefs to technical note

ariordan-redhat:
Cross references to the technical note module were failing because they used the module title instead of the module ID.
Affects `/titles/aap-on-gcp`.


ianf77 (approved):
LGTM

PR #241, opened 2023-07-25
Update proc-aws-marketplace.adoc

chrisahl:
typo.  100 not 110


chrisahl:
@ianf77 @ariordan-redhat I assume I need to submit a duplicate PR for aap-clouds-2.3 in addition to aap-clouds-latest?

PR #251, opened 2023-07-27
Fix con-tech-note-linux-files-owned-by-root file name

thenets:


stories/topics/con-tech-note-linux-files-owned-by-root.adoc
```diff
@@ -1,4 +1,4 @@
-[id="tech-note-linux-files-owned-by-root"]
+[id="con-tech-note-linux-files-owned-by-root"]
```
ianf77, line 1: If you change this, you'll have to change all the cross references to it. I changed all the cross references to match the original title a few days ago as it was breaking the build. (That's changing the ID, not the title)

ianf77 (commented):
One comment

ianf77 (approved):
LGTM

PR #264, opened 2023-08-14
Created image attribute

ianf77:
To replace image statements

DOC: Create new attribute for IMAGE value

https://issues.redhat.com/browse/AAP-14901

### DIFF
--- a/attributes/attributes.adoc
+++ b/attributes/attributes.adoc
@@ -18,6 +18,7 @@
 :GCP: Google Cloud Platform
 :AAPonAWS: Ansible Automation Platform from AWS Marketplace
 :AAPonGCP: Ansible Automation Platform from GCP Marketplace
+:ImageName: 2.4.20230630
 
 // Automation Mesh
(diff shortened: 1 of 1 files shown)


chrisahl (changes requested):
Several comments:
- I think we want the whole name as an attribute, for example: 
registry.redhat.io/ansible-on-clouds/ansible-on-clouds-ops-rhel8:2.2.20230215
because other parts of the name can change.  For example, in this last release, we changed from rhel8 to rhel9.

- I see this is targeted to branch aap-clouds-latest?  I thought we would be doing this work on the next release, not the current/latest one?  So whatever we name the next release branch, we should do the work in that branch.  

- Would like to verify that using this attribute in the bash examples shows up properly when rendered.  Would these changes be included in the same PR?

chrisahl:
ok, makes sense.  So I would say change the attribute key to :OpImageName: as it is for the Operational image and use the full value as shown previously.

chrisahl (approved):

PR #267, opened 2023-08-15
Replace image names and registry.redhat.io witth attributes

ianf77:
Replace with attributes


stories/topics/proc-aws-from-stack-pull-container-image.adoc
```diff
@@ -7,24 +7,25 @@
 +
 [NOTE]
 ====
-Before pulling the docker image, ensure you are logged in to registry.redhat.io using docker. Use the following command to login to registry.redhat.io.
+Before pulling the docker image, ensure you are logged in to {Registry} using docker. Use the following command to login to {Registry}.
```
ariordan-redhat, line 10: @ianf77 Is this procedure used in any assemblies?

ianf77: Possibly not, but it wasn't archived.
There's a gcp file proc-gcp-pull-backup-container-image which seems to have the wrong image. Either they missed it or it's not used. I can't find the value in the docs.

stories/topics/tech-note-aws-upgrade.adoc
```diff
@@ -2,4 +2,4 @@
...
\ No newline at end of file
+Upgrading {PlatformNameShort} to {ImageRef} updates the protocol of the listeners attached to the internal load balancers from HTTP to HTTPs. To update the listener protocol, the listener must be deleted and recreated properly. For this operation to succeed there must not have been an added resource dependancy on the listener. If there is a resource dependancy on the listener, this halts the upgrade, and the dependancy must be manually removed for the upgrade to succeed. When the upgrade has suceeded, the dependant resource can be recreated. You must revalidate any additional added networking configurations after upgrade.
```
ariordan-redhat, line 5:
```suggestion
Upgrading {PlatformNameShort} to {ImageRef} updates the protocol of the listeners attached to the internal load balancers from HTTP to HTTPs.
To update the listener protocol, the listener must be deleted and recreated properly.
For this operation to succeed there must not have been an added resource dependency on the listener.
If there is a resource dependency on the listener, this halts the upgrade, and the dependency must be manually removed for the upgrade to succeed.
When the upgrade has succeeded, the dependent resource can be recreated.
You must revalidate any additional added networking configurations after upgrade.
```

PR #277, opened 2023-08-21
2.4 Replace image names and registry.redhat.io with attributes (#267)

ianf77:
* Replace image names and registry.redhat.io witth attributes


ariordan-redhat (approved):
LGTM

PR #280, opened 2023-08-22
Additional instruction added

ianf77:
Additional step after modifying fstab

[DDF] Need to execute `systemctl daemon-reload` after you modify the fstab

https://issues.redhat.com/browse/AAP-15204


EMcWhinn (approved):
LGTM

PR #295, opened 2023-08-25
Updating route for webhook rulebooks

EMcWhinn:
Updating route to be HTTPS and job name to append the port

Route needs updating to properly work for EDA

https://issues.redhat.com/browse/AAP-15433

Affects `titles/eda-user-guide`


ianf77 (approved):
LGTM

PR #307, opened 2023-08-29
Update ref-architecture.adoc

ariordan-redhat:
Fix syntax in AAP hardening guide module


ariordan-redhat:
Content incorporated into 2.4 in another PR, so no backport PR needed.

PR #320, opened 2023-08-31
Add context to xref in aap-hardening module

ariordan-redhat:
Fix an xref in an aap-hardening module: In the xref, include the context from the assembly that contains the module that's referenced.


ariordan-redhat:
PR incorporated into 2.4 in another PR, so no backport PR.

PR #376, opened 2023-09-08
Create section 20 "Supported Inventory" of the Controller UG

EMcWhinn:
Added assembly and reference modules

Create Section 20 of the Controller user Guide

https://issues.redhat.com/browse/AAP-15831

### DIFF
--- a/downstream/modules/platform/ref-controller-amazon-web-services.adoc
+++ b/downstream/modules/platform/ref-controller-amazon-web-services.adoc
@@ -2,6 +2,7 @@
 
 = Amazon Web Services EC2
 
+[literal, options="nowrap" subs="+attributes"]
 ----
 compose:
   ansible_host: public_ip_address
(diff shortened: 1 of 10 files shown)


downstream/modules/platform/ref-controller-amazon-web-services.adoc
```diff
@@ -0,0 +1,109 @@
+[id="controller-amazon-web-services"]
+
+= Amazon Web Services EC2
+
+----
```
ianf77, line 5: Can you add a code block header before each [literal, options="nowrap" subs="+attributes"]

It allows you to include attributes and avoid the copy/paste copying any $ or # signs.

ianf77 (commented):
One comment

downstream/modules/platform/ref-controller-amazon-web-services.adoc
EMcWhinn: Done

ianf77 (approved):
LGTM

PR #380, opened 2023-09-08
Create section 20 "Supported Inventory" of the Controller UG (#376)

EMcWhinn:
* Create section 20 "Supported Inventory" of the Controller UG

Added assembly and reference modules

Create Section 20 of the Controller user Guide

https://issues.redhat.com/browse/AAP-15831

* Adding block headers


ianf77 (approved):
LGTM

PR #382, opened 2023-09-08
Delete temporary file

ariordan-redhat:
Delete temporary file so that `titles/repo-management` can be deleted.


dcdacosta (approved):
LGTM!

PR #383, opened 2023-09-08
AAP15779 - adding missing assembly

dcdacosta:
This PR includes an assembly that was missed in the Getting started guide and updates some leveloffsets to correct topic nesting.


hherbly (approved):
LGTM!

PR #466, opened 2023-09-21
Add modules to self-managed azure framework

ianf77:
Add modules

Create framework for self-managed Azure

https://issues.redhat.com/browse/AAP-16221


ariordan-redhat (approved):
LGTM

PR #477, opened 2023-09-25
Corrected statement about special characters in passwords

ianf77:
Correction

Clarify what characters are not acceptable for passwords

https://issues.redhat.com/browse/AAP-15242


AlanCoding (approved):
Yeah, this is much better, and at least gives something usable. While I'm sure we could allow many more, it's difficult to manage that testing burden, and if QE isn't already aligned to test it, I don't want to make such a recommendation.

ariordan-redhat (approved):
LGTM

PR #481, opened 2023-09-26
Correct variable typo for automationhub_pg_password

michellemacrh:
Correct typo with variable from automation_pg_password to automationhub_pg_password

[DDF] I'm not sure this is the correct parameter to use? Should it not be automationHUB_pg_password?

https://issues.redhat.com/browse/AAP-16095


EMcWhinn (approved):
LGTM

PR #513, opened 2023-09-28
Added note about ISP blocking port 445

rogrange:

PR #514, opened 2023-09-28
Added note about ISP blocking port 445 (#513)

rogrange:
* Added note about ISP blocking port 445

* Update ref-accessing-control-auto-hub-eda-control.adoc

* Update ref-accessing-control-auto-hub-eda-control.adoc


sayjadha (commented):
LGTM.

PR #532, opened 2023-10-02
Corrected statement about special characters in passwords (#477)

ianf77:
Clarify what characters are not acceptable for passwords https://issues.redhat.com/browse/AAP-15242


ariordan-redhat (approved):
LGTM

PR #291, opened 2023-08-23
AAP-11096 Add procedures changing SSL certs and keys

ariordan-redhat:

PR #553, opened 2023-10-06
2.3 AAP-11096 Add changing SSL certs and keys chapter to operations g…

ariordan-redhat:
…uide (#291)

AAP-11096
Add a chapter to the Operations guide (downstream/titles/aap-operations-guide) that describes how to replace SSL certs and keys.


hherbly (approved):
LGTM!

PR #599, opened 2023-10-16
Attribute for docs feedback in installation guide

ariordan-redhat:
Create an attribute for the documentation feedback assembly so that we only have to change an attribute in a branch rather than all the master.adoc files when feedback policies are updated.

Affects `titles/aap-installation-guide`.

Backport to 2.4 only: earlier versions do not include DDF.


ianf77 (approved):
LGTM

ianf77 (approved):

PR #612, opened 2023-10-17
Break out code blocks for easier copying of config details

michellemacrh:
Separate one code block into two to make it easier to copy relevant information

[DDF] This should be a separate code block so you can actually copy paste it in your terminal.

https://issues.redhat.com/browse/AAP-16743


ariordan-redhat (approved):
LGTM

PR #625, opened 2023-10-18
Adjust machine type and minimum count

dhageman:
We have changed the machine type and the minimum count for AAP on Azure.  This pull request adjusts the documentation to reflect the current state of the offer.


ariordan-redhat (approved):
LGTM

PR #633, opened 2023-10-19
Corrections to Controller UG

ianf77:
As a result of sanity check

Sanity check completed Controller User Guide

https://issues.redhat.com/browse/AAP-17315


ariordan-redhat (approved):
LGTM

PR #638, opened 2023-10-19
Fix inverted commas and define context value

ariordan-redhat:
Define a value for `context` in master.adoc for the Lightspeed user guide, so that assemblies in the doc that have `context` in their IDs can be interpreted.
Affects `lightspeed/lightspeed-user-guide`


hherbly (approved):
LGTM, thank you!!

PR #660, opened 2023-10-24
Use attribute for feedback module (#599)

ariordan-redhat:
Backports #599 

Create an attribute for the documentation feedback assembly so that we only have to change an attribute in a branch rather than all the master.adoc files when feedback policies are updated.

Affects `titles/aap-installation-guide`.


ianf77 (approved):
LGTM

PR #685, opened 2023-10-25
AAP-15899 Updated EDA controller scenario in installation guide

jself-sudoku:
- Corrected the code sample, per advice from SMEs in jira comments.
- Also, removed the Important admonition stating, "Automation controller must be installed before you populate the inventory file with the following Event-Driven Ansible variables." Per SME, a partner, and a customer, this is not true. We were told to add this note months ago.

### DIFF
--- a/downstream/modules/platform/ref-single-eda-controller-with-internal-db.adoc
+++ b/downstream/modules/platform/ref-single-eda-controller-with-internal-db.adoc
@@ -16,7 +16,6 @@ Use this example to populate the inventory file to install {EDAcontroller}. This
 [automationedacontroller]
 automationedacontroller.example.com ansible_connection=local
 
-
 [all:vars]
 automationedacontroller_admin_password='<eda-password>'
 
--- a/downstream/modules/platform/ref-single-eda-controller-with-internal-db.adoc
+++ b/downstream/modules/platform/ref-single-eda-controller-with-internal-db.adoc
@@ -7,17 +7,13 @@ Use this example to populate the inventory file to install {EDAcontroller}. This
 
 [IMPORTANT]
 ====
-{ControllerNameStart} must be installed before you populate the inventory file with the following {EDAName} variables.
-====
-
-[IMPORTANT]
-====
 {EDAcontroller} must be installed on a separate server and cannot be installed on the same host as {HubName} and {ControllerName}.
 ====
 -----
 # Automation EDA Controller Configuration
 #
 
+[all:vars]
 automationedacontroller_admin_password='<eda-password>'
 
 automationedacontroller_pg_host=''
@@ -27,17 +23,9 @@ automationedacontroller_pg_database='automationedacontroller'
 automationedacontroller_pg_username='automationedacontroller'
 automationedacontroller_pg_password='<password>'
 
-# The full routable URL used by EDA to connect to a controller host.
-# This URL is required if there is no Automation Controller configured
-# in inventory.
-#
 automation_controller_main_url = 'https://controller.example.com/'
  
-# Boolean flag used to verify Automation Controller's
-# web certificates when making calls from Automation EDA Controller.
-#
 automationedacontroller_controller_verify_ssl = true
-
 -----
 
 [role="_additional-resources"]
--- a/downstream/modules/platform/ref-single-eda-controller-with-internal-db.adoc
+++ b/downstream/modules/platform/ref-single-eda-controller-with-internal-db.adoc
@@ -23,9 +23,17 @@ automationedacontroller_pg_database='automationedacontroller'
 automationedacontroller_pg_username='automationedacontroller'
 automationedacontroller_pg_password='<password>'
 
+# The full routable URL used by EDA to connect to a controller host.
+# This URL is required if there is no Automation Controller configured
+# in inventory.
+#
 automation_controller_main_url = 'https://controller.example.com/'
  
+# Boolean flag used to verify Automation Controller's
+# web certificates when making calls from Automation EDA Controller.
+#
 automationedacontroller_controller_verify_ssl = true
+
 -----
 
 [role="_additional-resources"]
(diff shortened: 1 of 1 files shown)


downstream/modules/platform/ref-single-eda-controller-with-internal-db.adoc
```diff
@@ -27,17 +23,9 @@ automationedacontroller_pg_database='automationedacontroller'
 automationedacontroller_pg_username='automationedacontroller'
 automationedacontroller_pg_password='<password>'
 
-# The full routable URL used by EDA to connect to a controller host.
```
Alex-Izquierdo, line 30: Hi @jamesmarshall24 @jself-sudoku I may missing context. Why we remove these comments? it looks like a valid info for the customer.

jamesmarshall24: @Alex-Izquierdo I don't really know do we want to keep them in or not? It seems like most of the examples on this page leave out comments from the inventory file while others keep them.

jamesmarshall24: @jself-sudoku I would say this is your call. I do think maybe we should keep them in to give context to the variables. I'll give a follow up suggestion change to what Alejandro asked.

downstream/modules/platform/ref-single-eda-controller-with-internal-db.adoc
```diff
@@ -27,17 +23,9 @@ automationedacontroller_pg_database='automationedacontroller'
...
-# This URL is required if there is no Automation Controller configured
-# in inventory.
-#
 automation_controller_main_url = 'https://controller.example.com/'
```
jamesmarshall24, line 26:
```suggestion
# The full routable URL used by EDA to connect to a controller host.
# This URL is required if there is no Automation Controller configured
# in inventory.
#
automation_controller_main_url = 'https://controller.example.com/'

# Boolean flag used to verify Automation Controller's
# web certificates when making calls from Automation EDA Controller.
#
automationedacontroller_controller_verify_ssl = true
```

jamesmarshall24 (changes requested):

downstream/modules/platform/ref-single-eda-controller-with-internal-db.adoc
jself-sudoku: @jamesmarshall24 and @Alex-Izquierdo - I only removed those to comply with the example that @jamesmarshall24 added to the comments in the jira. I can add them back with no problem. It's just a matter of undoing what was completed before. As for me making the call, I don't feel comfortable doing that at this point. I'm hoping to gain more knowledge as we progress, but I'm definitely not there yet. So, I'll pass this through to you both after I've added those lines back.

downstream/modules/platform/ref-single-eda-controller-with-internal-db.adoc
jself-sudoku: @jamesmarshall24 - I assumed you meant that I should remove this entire line. Let me know if you intended something different.

jself-sudoku: I have updated this PR based on suggested changes.

jamesmarshall24: @jself-sudoku the way git shows diffs can be confusing, but the `automation_controller_main_url` should still be there just moved into the section under the comments. If you look at the green in my suggested change, you'll still see it there. Can you re-do the PR with that line added back?

It also looks like we're missing this line too at the top:
```
[automationedacontroller]
automationedacontroller.example.com ansible_connection=local
```

jself-sudoku: @jamesmarshall24 - can you use the **Files changed** tab to show where these lines should be? 

Here's what's confusing . I used the exact example that displays in the published information and just added the `[all: vars]` above automationedacontroller_admin_password='<eda-password>'

If there are missing lines, they've been missing since the beginning. At this point, I'm not sure where you're saying to add the lines that you say are missing now. Should it be the following (missing text in **bold**)? 

# Automation EDA Controller Configuration
#

**[automationedacontroller]
automationedacontroller.example.com ansible_connection=local**

[all:vars]
automationedacontroller_admin_password='<eda-password>'

automationedacontroller_pg_host=''
automationedacontroller_pg_port=5432

automationedacontroller_pg_database='automationedacontroller'
automationedacontroller_pg_username='automationedacontroller'
automationedacontroller_pg_password='<password>'

jself-sudoku:
@jamesmarshall24  and @Alex-Izquierdo - I have added the lines that James requested at the top of the code example. Please note, these lines were not in the original code example. So we added them, per James' request. Also, we added back the **automation_controller_main_url = 'https://controller.example.com/'** that was missing from the code sample that James provided initially in this PR as an example for resolving this issue. The [all:vars] was added initially. So I think that captures all that needs to be updated or restored. 
 
 Please confirm that the latest updated code section is exactly what the customer needs to clear any confusion. If possible, can I get technical approval on this by the end of today, Monday, 10/30? Thanks in advance.

downstream/modules/platform/ref-single-eda-controller-with-internal-db.adoc
jamesmarshall24: I think we also need this still at the bottom:
```
# This URL is required if there is no Automation Controller configured
# in inventory.
#
automation_controller_main_url = 'https://controller.example.com/'

# Boolean flag used to verify Automation Controller's
# web certificates when making calls from Automation EDA Controller.
#
automationedacontroller_controller_verify_ssl = true
```

Alex-Izquierdo (approved):
lgtm

jself-sudoku:
Thanks, @Alex-Izquierdo.

jamesmarshall24 (approved):

jself-sudoku:
Thanks for the approval, @jamesmarshall24 .

hherbly (approved):
LGTM!

PR #696, opened 2023-10-26
Update proc-create-credential.adoc

annibond:
Added final statement regarding the creation of a credential. At least one of three fields is mandatory, but it's also possible to create one in the web UI without specifying. The reason is because it's specified to the user creating it.


rogrange:
I am a little confused since I do not know the context, but this is a bit awkward. 

instead of "Filling in at least one of the fields *organization*, *user*, or *team* is mandatory for creating a credential. It is possible to create a credential in the user interface. This is possible, because your credential links to your user profile."

Maybe something like:  Filling in at least one of the fields *organization*, *user*, or *team* is mandatory, and can be done through the user interface." 

Is it necessary to add that  :This is possible, because your credential links to your user profile?" Would that not be evident?

annibond:
That is a much better wording than what I had. And I agree with your second comment. Going to implement those suggestions.

PR #718, opened 2023-10-31
Create section 11 of Controller Admin guide (#717)

ianf77:
Added assembly

Create section 11 of the Controller admin guide

https://issues.redhat.com/browse/AAP-17266


michellemacrh:
LGTM

PR #721, opened 2023-11-01
Create section 12 of the Controller Admin Guide

ianf77:
Added assembly and modules

Create section 12 of the Controller admin guide

https://issues.redhat.com/browse/AAP-17267


annibond (approved):
These changes all look good to me.

PR #736, opened 2023-11-06
Backport-2.4-725 (for AAP-14288) Added an admonition about creating a different name for EDA controller

jself-sudoku:
- Added "Important" admonition to the Deploying the Red Hat Ansible Automation Platform operator on OpenShift Container Platform, Chapter 6, informing customers that they should give their EDA controller a different name if they are deploying it in the same namespace as other AAP components.

- Also, per @rooftopcellist, added a new step to specify name for EDA controller that's distinct from OpenShift Container Platform namespace


dcdacosta (approved):
LGTM!

PR #754, opened 2023-11-10
Remove direct documentation feedback option

ariordan-redhat:
Remove direct documentation feedback module from the AAP 2.4 docs.
Replace it with the original feedback procedure from 2.2.

PR #762, opened 2023-11-13
2.4 Remove direct documentation feedback option

ariordan-redhat:
Remove direct documentation feedback module from the AAP 2.4 docs.
Replace it with the original feedback procedure from 2.2.
Backports #754 to `2.4`.


ianf77 (approved):
LGTM

PR #769, opened 2023-11-14
Remove Operator related items from bundle installer release notes (#768)

michellemacrh:
Remove Operator related release items from bundle installer release notes. These will be added at a later date in their own section. https://issues.redhat.com/browse/AAP-18174


ariordan-redhat (approved):
LGTM

PR #774, opened 2023-11-14
Remove DDF from clouds docs

ariordan-redhat:
Remove the Direct Document Feedback process from the clouds docs.
Backport to aap-clouds-2.4.


ianf77 (approved):
LGTM

PR #776, opened 2023-11-14
aap-clouds-2.4 Remove DDF from clouds docs (#774)

ariordan-redhat:
Remove the Direct Document Feedback process from the clouds docs.
Backports #774 to `aap-clouds-2.4`.


ianf77 (approved):
LGTM

PR #790, opened 2023-11-16
Made editorial changes to chapter 4 disconnected installation.

rogrange:


downstream/modules/platform/con-aap-installation-on-disconnected-rhel.adoc
```diff
@@ -4,22 +4,24 @@
...
+You can install {PlatformNameShort} {ControllerName} and a {PrivateHubName} without an internet connection by using the installer-managed database located on the {ControllerName}. The setup bundle is recommended for disconnected installation because it includes additional components that make installing {PlatformNameShort} easier in a disconnected environment. These include the {PlatformNameShort} Red Hat package managers (RPMs) and the default {ExecEnvShort} (EE) images.
 
-== System Requirements
+== System requirements for disconnected installation
 
-Hardware requirements are documented in the link:https://access.redhat.com/documentation/en-us/red_hat_ansible_automation_platform/{PlatformVers}/html/red_hat_ansible_automation_platform_installation_guide/platform-system-requirements#ref-controller-system-requirements[{PlatformNameShort} Installation Guide].
+Ensure that your system has all the hardware requirements before performing a disconnected installation of Ansible Automation Platform. You can find these requirements in Chapter 2. System Requirements [link] in the {PlatformNameShort} Installation Guide..
```
dcdacosta, line 11: Remove the extra period here.

downstream/modules/platform/proc-installing-the-ansible-builder-rpm.adoc
```diff
@@ -2,19 +2,25 @@
...
 = Installing the ansible-builder RPM
 
-.Procedure
+[role="_abstract"]
 
-. On a RHEL system, install the ansible-builder RPM. This can be done in one of several ways:
+In order to create custom {ExecEnvShort} images in a disconnected environment, you will need to have a working copy of ansible-builder installed. On a RHEL system there are several ways to do this.
+
+* Subscribe the RHEL box to a Satellite on the disconnected network.
+* Attach the Ansible Automation Platform subscription and enable the AAP repo.
```
dcdacosta, line 10: AAP is not an approved acronym and I think it should be replaced here unless that is the actual name of the repo.

dcdacosta (approved):
@rogrange Two minor comments but I went ahead and merged since they're nits and can be handled when you return from PTO.

downstream/modules/platform/proc-installing-the-ansible-builder-rpm.adoc
rogrange: I don't know if that is the name of the repo or not. It was written like that by the devs, so I used it. I doubt there is anyone around now who can tell me if it is or not.

PR #825, opened 2023-11-29
Update variable name

ianf77:
Updated URL to hostname

[DDF] This should be `automation_hub_hostname` as the URL includes the protocol, including the protocol will produce an

https://issues.redhat.com/browse/AAP-18024


ariordan-redhat (approved):
LGTM

PR #833, opened 2023-11-30
Add additional information and link for subscription compliance

ianf77:
Added text

Add information to Subscription section

https://issues.redhat.com/browse/AAP-17588


tvo318 (commented):
I think this is fine, as long as your style guidelines allow referencing other content.

PR #830, opened 2023-11-29
Second sanity check of Controller Admin Guide part 2

EMcWhinn:
https://issues.redhat.com/browse/AAP-18518

Affects titles/controller-admin-guide

PR #834, opened 2023-11-30
Second sanity check of Controller Admin Guide part 2 (#830)

EMcWhinn:
* Second sanity check of Controller Admin Guide part 2

https://issues.redhat.com/browse/AAP-18518

Affects titles/controller-admin-guide

* Second commit, second sanity check part 2

* Peer review edits


ianf77 (approved):
LGTM

PR #855, opened 2023-12-01
AAP-18404: Added info. about removing licenses for deactivated users

sayjadha:
This PR addresses JIRA https://issues.redhat.com/browse/AAP-18404.

Changes:

- Added info. about removing Lightspeed licenses for deactivated RH-SSO users.

### DIFF
--- a/lightspeed/modules/proc_remove-seat-licenses.adoc
+++ b/lightspeed/modules/proc_remove-seat-licenses.adoc
@@ -5,10 +5,14 @@
 = Removing seat licenses
 
 [role="_abstract"]
-You can remove users who no longer need to access {LightspeedShortName}, which makes licenses available for others to use.
+You can remove users who no longer need to access {LightspeedShortName}, which makes licenses available for others to use. 
+
+You can also remove seat licenses of deactivated RH-SSO users, so that they can no longer access {LightspeedShortName}. When RH-SSO users are deactivated, their {LightspeedShortName} seat licenses are not removed automatically. These users can continue accessing {LightspeedShortName} until the next time they are forced to reauthenticate with RH-SSO. To prevent deactivated RH-SSO users from accessing {LightspeedShortName}, you must remove their {LightspeedShortName} seat licenses.
+
 [NOTE]
 ====
-This task removes the Ansible user's {LightspeedShortName} seat license only; it does not remove the user from your Red Hat Customer Portal organization.
+When you remove a user, the Ansible user's {LightspeedShortName} seat license only is removed; it does not remove the user from your Red Hat Customer Portal organization. 
+
 ====
 
 .Procedure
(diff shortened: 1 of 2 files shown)


lightspeed/modules/proc_remove-seat-licenses.adoc
```diff
@@ -5,10 +5,14 @@
 = Removing seat licenses
 
 [role="_abstract"]
-You can remove users who no longer need to access {LightspeedShortName}, which makes licenses available for others to use.
+You can remove users who no longer need to access {LightspeedShortName}, which makes licenses available for others to use. 
+
+You can also remove seat licenses of deactivated RH-SSO users, so that they can no longer access {LightspeedShortName}. When RH-SSO users are deactivated, their {LightspeedShortName} seat licenses are not removed automatically. These users can continue accessing {LightspeedShortName} until the next time they are forced to reauthenticate with RH-SSO. To prevent deactivated RH-SSO users from accessing {LightspeedShortName}, you must remove their {LightspeedShortName} seat licenses.
```
robinbobbitt, line 10: This looks good but I wonder if we should say "You should" instead of "You can".

robinbobbitt: Thank you for the explanation. I still can't get comfortable with the "You can" in the first sentence. Given the sentence before it, the first half of the sentence feels like we're offering up another way to get more licenses freed up for other users, and if I don't care about that at the moment, I will stop reading. Maybe we should just delete the first sentence entirely and replace it with "IMPORTANT: " or something?

sayjadha: I understand your point, and have made minor edits. Could you please check if the current wording works? Thank you.

sayjadha: Incorporated additional updates per Slack convo., thank you.

lightspeed/modules/proc_remove-seat-licenses.adoc
```diff
@@ -5,10 +5,12 @@
 = Removing seat licenses
 
 [role="_abstract"]
-You can remove users who no longer need to access {LightspeedShortName}, which makes licenses available for others to use.
-[NOTE]
+Remove users who no longer need to access {LightspeedShortName}, which makes licenses available for others to use. When you remove a user, the Ansible user's {LightspeedShortName} seat license only is removed; it does not remove the user from your Red Hat Customer Portal organization.
```
robinbobbitt, line 8: With this "You can" gone, the sentence doesn't really read right. How about: "You can remove seat assignments from users who no longer need to access {LightspeedShortName}, making those seats available for others to use. When you remove a user from the Seat Management tool, the Ansible user's {LightspeedShortName} seat license only is removed; it does not remove the user from your Red Hat Customer Portal organization."

robinbobbitt (approved):

lightspeed/modules/proc_remove-seat-licenses.adoc
sayjadha: Readded 'you can', rest text is the same. TY.

PR #878, opened 2023-12-06
AAP-14712 Add content to upgrade/migration guide for disabling activations for EDA

jself-sudoku:
Added a new discrete heading to [Chapter 2.1 in the RH AAP Upgrade and Migration Guide](https://access.redhat.com/documentation/en-us/red_hat_ansible_automation_platform/2.4/html-single/red_hat_ansible_automation_platform_upgrade_and_migration_guide/index#aap-upgrade-planning_aap-upgrading-platform) that informs the customer to stop all activations before beginning the upgrade process if they plan to deploy EDA, per proposal from SMEs.

### DIFF
--- a/downstream/modules/platform/con-aap-upgrade-planning.adoc
+++ b/downstream/modules/platform/con-aap-upgrade-planning.adoc
@@ -33,4 +33,4 @@ Before you begin the upgrade process, review the following considerations to pla
 == {EDAcontroller}
 //ATTENTION: Remove this section for EDA 1.0.4; customers will no longer need to perform deactivation because services will be automatically restored after upgrade and migration. 
 
-* If you are currently running {EDAcontroller} and plan to deploy it when you upgrade, it is recommended that you deactivate all current activations before upgrading to {PlatformNameShort} {PlatformVers} to ensure that only new activations run after the upgrade process has completed. This prevents possibilities of orphaned containers running activations from the previous version.
+* If you are currently running {EDAcontroller} and plan to deploy it when you upgrade, it is recommended that you disable all current activations before upgrading to {PlatformNameShort} {PlatformVers} to ensure that only new activations run after the upgrade process has completed. This prevents possibilities of orphaned containers running activations from the previous version.
--- a/downstream/modules/platform/con-aap-upgrade-planning.adoc
+++ b/downstream/modules/platform/con-aap-upgrade-planning.adoc
@@ -33,4 +33,4 @@ Before you begin the upgrade process, review the following considerations to pla
 == {EDAcontroller}
 //ATTENTION: Remove this section for EDA 1.0.4; customers will no longer need to perform deactivation because services will be automatically restored after upgrade and migration. 
 
-* If you are currently running {EDAcontroller} and plan to deploy it when  you upgrade, it is recommended that you deactivate all current activations before upgrading to {PlatformNameShort} {PlatformVers} to ensure that only new activations run after the upgrade process has completed. This prevents possibilities of orphaned containers running activations from the previous version.
+* If you are currently running {EDAcontroller} and plan to deploy it when you upgrade, it is recommended that you deactivate all current activations before upgrading to {PlatformNameShort} {PlatformVers} to ensure that only new activations run after the upgrade process has completed. This prevents possibilities of orphaned containers running activations from the previous version.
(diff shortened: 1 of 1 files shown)


Alex-Izquierdo (approved):
The content looks good, just I have a question due to my lack of knowledge about how the docs portal works. Are we sure this text will be displayed only for upgrades to <=1.0.3?

jself-sudoku:
> The content looks good, just I have a question due to my lack of knowledge about how the docs portal works. Are we sure this text will be displayed only for upgrades to <=1.0.3?

This is a great question, @Alex-Izquierdo . I actually wondered about this based on the [EDA Upgrade Process Google doc](https://docs.google.com/document/d/1Tgd1bMXmngbOjf8aJdLBLeHsz7y4aOH9vkyI7pZMBXI/edit#heading=h.m97y9asqbd8u) you included as a draft in jira https://issues.redhat.com/browse/AAP-14712.

So should I propose a different solutions that break things down by version as you have things in the google doc? I'm going to include further comments in the jira.

Alex-Izquierdo:
@jself-sudoku as it is now is fine and 1.0.3 is our latest version. This message is valid for any upgrade until 1.0.3. 

My question is more thinking in the future, for 1.0.4 we expect to remove this advice, so we should take care of it and be sure it is displayed only for the right versions. I don't know if we are taking care of intermediate versions or it is always the same for the latest 2.4. In that case, it may be just a matter to remove this when 1.0.4 is released.

jself-sudoku:
Oh, okay. I see. Yes, we can remove it for 1.0.4 if that is released with AAP 2.5. What I can do is include a commented out statement that let's us know to remove the section with the next release, which would be 2.5.

downstream/modules/platform/con-aap-upgrade-planning.adoc
```diff
@@ -28,3 +28,9 @@ Before you begin the upgrade process, review the following considerations to pla
 [role="_additional-resources"]
 .Additional resources
 * <<editing-inventory-file-for-updates_{context}, Setting up the inventory file >>
+
+[discrete]
+== {EDAcontroller}
+//ATTENTION: Remove this section for AAP 2.5; customers will no longer need to perform deactivation because services will be automatically restored after upgrade and migration.
```
Alex-Izquierdo, line 34:
```suggestion
//ATTENTION: Remove this section for EDA 1.0.4; customers will no longer need to perform deactivation because services will be automatically restored after upgrade and migration. 
```

jself-sudoku: Okay. Duly noted. When will EDA 1.0.4 be released? Will that happen with AAP 2.5 or will it be done with separately?

jself-sudoku: Thanks for reviewing the update. The note was added for me or anyone else on the docs team. It won't be viewed by customers.

Alex-Izquierdo: 1.0.4 will be released in 2.4, thats why I think its better to point to eda 1.0.4 instead of aap 2.5

jself-sudoku: Got it! Thanks for the clarification.

downstream/modules/platform/con-aap-upgrade-planning.adoc
```diff
@@ -28,3 +28,9 @@ Before you begin the upgrade process, review the following considerations to pla
 [role="_additional-resources"]
 .Additional resources
 * <<editing-inventory-file-for-updates_{context}, Setting up the inventory file >>
+
+[discrete]
+== {EDAcontroller}
+//ATTENTION: Remove this section for EDA 1.0.4; customers will no longer need to perform deactivation because services will be automatically restored after upgrade and migration. 
+
+* If you are currently running {EDAcontroller} and plan to deploy it when you upgrade to {PlatformNameShort} {PlatformVers}, it is recommended that you disable all {EDAName} activations before upgrading  to ensure that only new activations run after the upgrade process has completed. This prevents possibilities of orphaned containers running activations from the previous version.
```
dcdacosta, line 36: It looks like there might be an extra space after "...before upgrading"

dcdacosta (approved):
One minor nit, otherwise, LGTM!

downstream/modules/platform/con-aap-upgrade-planning.adoc
jself-sudoku: Got it! Done.

dcdacosta (approved):
LGTM!

PR #885, opened 2023-12-07
PPAH Delete non-ppah content from PPAH branch

ariordan-redhat:
PPAH is a separate product from the other AAP products.
It is in a separate branch to the AAP releases and will therefore be published on a different splash page.

This PR deletes all content that is not related to PPAH.
It also puts ppah content into `ppah/` directories so that there's less chance of confusion about which files are necessary when git shows `untracked files` messages.


hherbly (approved):
LGTM!

PR #889, opened 2023-12-11
Corrected port definitions

ianf77:
Removed quotes

[DOC] Database port definition should match default inventory file

https://issues.redhat.com/browse/AAP-18919


ariordan-redhat (commented):
LGTM. One question, though: I took a look at the build and searched for `5432`.
There's one instance where it's still `'5432'`, in `/modules/platform/proc-installing-the-aap-setup-bundle.adoc`
Do you want to change that one also?

ariordan-redhat (approved):
LGTM

PR #890, opened 2023-12-11
Corrected port definitions (#889)

ianf77:
* Corrected port definitions: Removed quotes [DOC] Database port definition should match default inventory file https://issues.redhat.com/browse/AAP-18919


ariordan-redhat (approved):
LGTM

PR #891, opened 2023-12-11
PPAH Delete sync script from ppah-main branch

ariordan-redhat:
Delete the sync script from the `ppah-main` branch.
This branch is not published, so there is no need to sync it to Gitlab.


hherbly (commented):
LGTM.

PR #892, opened 2023-12-11
Remove PPAH from main downstream docs branch

ariordan-redhat:
Remove PPAH from main branch. It's a separate product, so it only exists in PPAH branches.


hherbly (commented):
LGTM.

PR #915, opened 2023-12-20
AAP-18088: adds info on download count configuration; fixes a namespace error

hherbly:
This PR adds info on setting a download count configuration for automation hub on OCP. The change will appear in the Deploying the RH AAP operator on OpenShift Container Platform guide. 

See[ AAP-18088](https://issues.redhat.com/browse/AAP-18088) for further context.

### DIFF
--- a/downstream/modules/platform/ref-ocp-additional-configs.adoc
+++ b/downstream/modules/platform/ref-ocp-additional-configs.adoc
@@ -10,9 +10,7 @@ A collection download count can help you understand collection usage. To add a c
 -----
 spec:
   pulp_settings:
-    api_root: "/api/galaxy/pulp/"
     ansible_collect_download_count: true 
-
 -----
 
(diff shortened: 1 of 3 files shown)


downstream/modules/platform/ref-ocp-additional-configs.adoc
```diff
@@ -0,0 +1,18 @@
...
+= Additional configurations
+
+[role="_abstract"]
+
+A collection download count can help you understand collection usage. To add a collection download count to {HubName}, set the following configuration:
+
+-----
+spec:
+  pulp_settings:
+    api_root: "/api/galaxy/pulp/"
```
rooftopcellist, line 13: There is a potential that this `api_root` setting's value may change in the future.  I think it might be better to omit that line from this example, the yaml will still be valid and demonstrate how to specify settings without it.  

Otherwise this change looks good to me!

rooftopcellist (approved):

dcdacosta (approved):
LGTM!

PR #921, opened 2024-01-04
Move template files to templates directory

michellemacrh:
Move template files to a new templates directory so they are separate from snippets

Automate addition of async releases to the release notes

https://issues.redhat.com/browse/AAP-17585


emurtoug (approved):
LGTM

PR #939, opened 2024-01-11
AAP-18875 Update containerized install guide after TP refresh

rogrange:
These updates reflect new development functionality

Partially resolves: AAP-18875
See also: ANSTRAT-246


annibond (approved):
These all look good and appear to be appropriate, more easily understandable/readable changes!

PR #943, opened 2024-01-12
AAP-18875 Update containerized install guide after TP refresh (#939)

rogrange:
These updates reflect new development functionality

Partially resolves: AAP-18875
See also: ANSTRAT-246


dcdacosta (approved):
LGTM!

PR #980, opened 2024-01-26
AAP-17700: Further updates to Builder documentation

hherbly:
See [AAP-17700](https://issues.redhat.com/browse/AAP-17700) for context.

### DIFF
--- a/downstream/modules/builder/ref-scenario-using-authentication-ee.adoc
+++ b/downstream/modules/builder/ref-scenario-using-authentication-ee.adoc
@@ -22,14 +22,12 @@ additional_build_files:
 
 additional_build_steps:
   prepend_galaxy:
-    # Copy ansible.cfg from build directory to EE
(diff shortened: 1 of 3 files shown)


downstream/modules/builder/ref-scenario-using-authentication-ee.adoc
```diff
@@ -22,14 +22,12 @@ additional_build_files:
```
nitzmahone, line 22: `additional_build_files` isn't necessary at all for this one now, and unless you want it to be a complete example (in which case it needs `version` and `images`, `options` is also unnecessary.

nitzmahone (commented):
Yep, other than `additional_build_files` (and possibly `options`) being superfluous (for an "additive" sample anyway), that looks right now. Thanks!

jself-sudoku:
LGTM

PR #996, opened 2024-02-06
Correct mesh examples (#995)

ianf77:
Mesh Document example is not correct it is missing information. https://issues.redhat.com/browse/AAP-20181


ariordan-redhat (approved):
LGTM

PR #1016, opened 2024-02-16
Update assembly-network-ports-protocols.adoc

annibond:
Added one line as a note before the diagram about the hop node connecting execution and control nodes with the justification behind using them per Jira ticket AAP-17313


hherbly (approved):
LGTM!

PR #1018, opened 2024-02-16
Update proc-push-container.adoc

annibond:
The previous version of this doc has the user enter both their username and password into the CLI in step 1.

As per https://issues.redhat.com/browse/AAP-18025, this exposes the password to the shell history. It is more secure to allow Podman to prompt the user for the password.


hherbly (approved):
LGTM

PR #1032, opened 2024-02-21
Update docs for workflow and job template limits

EMcWhinn:
Adding clarity around limits for workflow and job templates

Workflow limit not working as expected

https://issues.redhat.com/browse/AAP-15512


g-murray:
LGTM

PR #1037, opened 2024-02-22
Replace plain text with attribute for content in the builder module (…

michellemacrh:
…#1036)

Replace plain text with attribute for content in the builder module. Affects `/titles/builder` and `/titles/builder/builder`.


emurtoug (approved):
LGTM

PR #1042, opened 2024-02-26
Replace plain text with attributes for content in the core module (#1…

michellemacrh:
…041)

Replace plain text with attributes for content in the core module. Affects `/titles/dev-guide/core`


g-murray:
LGTM

PR #1044, opened 2024-02-27
Replace plain text with attributes for content in the dev-guide module

michellemacrh:
Replace plain text with attributes for content in the dev-guide module.

Replace plain-text with attributes where appropriate

Affects the following titles:

- ./aap-operations-guide/platform/dev-guide
- ./upgrade/platform/dev-guide
- ./aap-operator-installation/platform/dev-guide
- ./aap-installation-guide/platform/dev-guide
- ./aap-planning-guide/platform/dev-guide
- ./automation-mesh/platform/dev-guide
- ./ocp_performance_guide/platform/dev-guide
- ./controller/controller-user-guide/platform/dev-guide
- ./controller/controller-admin-guide/platform/dev-guide
- ./controller/controller-getting-started/platform/dev-guide
- ./aap-operator-backup/platform/dev-guide
- ./aap-containerized-install/platform/dev-guide
- ./dev-guide/dev-guide

https://issues.redhat.com/browse/AAP-19894

PR #1045, opened 2024-02-27
Replace plain text with attributes for content in the dev-guide modul…

michellemacrh:
…e (#1044)

Replace plain text with attributes for content in the dev-guide module. https://issues.redhat.com/browse/AAP-19894


ariordan-redhat (commented):
LGTM

PR #1047, opened 2024-02-28
Add openssl commands to validate SSL certs (#1046)

michellemacrh:
Add openssl commands to validate SSL certs

Need to include openssl commands to validate the custom SSL certificates for AAP/PAH

https://issues.redhat.com/browse/AAP-20776


emurtoug (approved):
LGTM

PR #1053, opened 2024-02-29
Added information to inventory variable.

ianf77:
doc update inventory file variables

https://issues.redhat.com/browse/AAP-21254


downstream/modules/platform/ref-hub-variables.adoc
```diff
@@ -201,6 +201,11 @@ Generating a new token will invalidate the existing token.
 Default = 63072000 seconds, or two years.
 | *`nginx_tls_protocols`* | Defines support for `ssl_protocols` in Nginx.
 
+Values available `TLSv1.2`, `TLSv1.3`
```
dsavineau, line 204: More values are available but are considered weak and unsafe.
However, one could decide to enable them anyway if needed.

https://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_protocols

downstream/modules/platform/ref-hub-variables.adoc
```diff
@@ -201,6 +201,11 @@ Generating a new token will invalidate the existing token.
...
+
+If `nginx_tls-protocols` is set to `TLSv1.3` only TLSv1.3  is enabled.
```
dsavineau, line 206:
```suggestion
If `nginx_tls_protocols` is set to `TLSv1.3` only TLSv1.3  is enabled.
```

Note that for a single value like `TLSv1.3` then the `nginx_tls_protocols` variable is still a list so it should be set like `nginx_tls_protocols=['TLSv1.3']`

or in yaml format:
```yaml
nginx_tls_protocols:
  - TLSv1.3
```

downstream/modules/platform/ref-hub-variables.adoc
```diff
@@ -201,6 +201,15 @@ Generating a new token will invalidate the existing token.
 Default = 63072000 seconds, or two years.
 | *`nginx_tls_protocols`* | Defines support for `ssl_protocols` in Nginx.
 
+Values available `TLSv1`, `TLSv1.1, `TLSv1.2`, `TLSv1.3`
+
+The TLSv1.1 and TLSv1.2 parameters only work when OpenSSL 1.0.1 or higher is used.
+
+The TLSv1.3 parameter only works when OpenSSL 1.1.1 or higher is used.
+
+If `nginx_tls-protocols = ['TLSv1.3']' only TLSv1.3  is enabled.
```
dsavineau, line 210:
```suggestion
If `nginx_tls_protocols = ['TLSv1.3']` only TLSv1.3  is enabled.
```

PR #1066, opened 2024-03-04
Replace plain text with attributes for content in the eda module

michellemacrh:
Affects the following titles:

`.downstream/titles/eda/eda-user-guide/eda`
`.downstream/titles/eda/eda-getting-started-guide/eda`


g-murray:
LGTM

PR #1067, opened 2024-03-04
Replace plain text with attributes for content in the eda module (#1066)

michellemacrh:
Affects the following titles:

`./eda-user-guide/eda`
`./eda-getting-started-guide/eda`


g-murray:
LGTM

PR #1159, opened 2024-03-22
Replace plain text with attributes for content in the builder module

michellemacrh:
This PR is specific to the 2.3 branch and version of our docs. The `main` PR is #1036.

Affects the following title: `builder`.


g-murray:
LGTM

PR #1173, opened 2024-03-25
AAP-21516 Update links to use BaseURL attribute in platform assemblies

ariordan-redhat:
AAP-21516 Update links to use `BaseURL` attribute in platform assemblies


sayjadha (commented):
LGTM.

PR #1176, opened 2024-03-25
2.4 AAP-21516 Update links to use BaseURL attribute in platform assemblie…

ariordan-redhat:
Backports #1173 to 2.4

AAP-21516 Update links to use BaseURL attribute in platform assemblies


michellemacrh:
LGTM

PR #1226, opened 2024-04-15
Update links to use BaseURL attribute in assemblies

ariordan-redhat:
AAP-21516 Update links to use `BaseURL` attribute in assemblies.


ianf77 (approved):
LGTM

PR #1227, opened 2024-04-15
2.4 Update links to use BaseURL attribute in assemblies (#1226)

ariordan-redhat:
Backports #1226 to 2.4
AAP-21516 Update links to use `BaseURL` attribute in assemblies.


ianf77 (approved):
LGTM

PR #1228, opened 2024-04-15
Update links to use BaseURL attribute in modules

ariordan-redhat:
AAP-21516 Update links to use BaseURL attribute in modules.


ianf77 (approved):
LGTM

PR #1230, opened 2024-04-15
2.4 Update links to use BaseURL attribute in modules (#1228)

ariordan-redhat:
AAP-21516 Update links to use the `BaseURL` attribute in modules.


ianf77 (approved):
LGTM

PR #1229, opened 2024-04-15
Update links to use PlatformVers attribute

ariordan-redhat:
Update links in the EDA, Hub, and Hardening guides to use the `PlatformVers` attribute instead of `2.4`.

PR #1235, opened 2024-04-16
2.4 Update links to use PlatformVers attribute (#1229)

ariordan-redhat:
Update links in the EDA, Hub, and Hardening guides to use the `PlatformVers` attribute instead of `2.4`.


michellemacrh:
LGTM

PR #1236, opened 2024-04-16
AAP-21516 Update URLs in Lightspeed docs to use BaseURL

ariordan-redhat:
AAP-21516 Update URLs in Lightspeed docs to use the `BaseURL` attribute


g-murray:
LGTM

PR #1247, opened 2024-04-18
Create Archive directory 

g-murray:


ariordan-redhat:
Adding tmp.adoc in directory because GitHub doesn't recognize diffs for empty directories

PR #1267, opened 2024-04-23
Correct release note title name

michellemacrh:
The doc title name should match exactly what's on Pantheon otherwise the publish will fail. Reverting the title back to how it was previously before implementing feedback.


g-murray:
LGTM

PR #1274, opened 2024-04-25
AAP-21516 Update links to use BaseURL attribute

ariordan-redhat:
AAP-21516 Update links to use BaseURL attribute


EMcWhinn (approved):
LGTM

PR #1279, opened 2024-04-26
Corrected API version

ianf77:
Changed awx value for correct value

API version specified in the documentation of AAP Operator installation is incorrect

https://issues.redhat.com/browse/AAP-22949


g-murray:
LGTM

PR #1281, opened 2024-04-26
Corrected API version (#1279)

ianf77:
Changed awx value for correct value

API version specified in the documentation of AAP Operator installation is incorrect

https://issues.redhat.com/browse/AAP-22949


michellemacrh:
LGTM

PR #1291, opened 2024-04-29
Update for 2.4 primary menu snippets

dcdacosta:
Once the unified platform is fully implemented, some of the menu selections will become tertiary. In order to simplify the updates, I have included snippets to cover those instances. While the content for 2.4 will be the same, they will become unique in the future releases. I've also updated the attributes file to group the menu selections effected by these instances and include comments for future reference.


ariordan-redhat (approved):
LGTM

PR #1294, opened 2024-04-30
Add NavigatorStart attribute and update Navigator attribute (#1293)

michellemacrh:
According to Official Red Hat product names list (OPL), the correct usage is "automation content navigator".

It would be helpful to set the attributes like this:

- :Navigator: automation content navigator
- :NavigatorStart: Automation content navigator

Additionally, the usage of these attributes should be reviewed throughout the docs set to ensure it's used correctly.

Update Navigator attribute and add NavigatorStart attribute

https://issues.redhat.com/browse/AAP-21433


g-murray:
LGTM

PR #1299, opened 2024-04-30
2.4 AAP-17190 Modules for devtools workflow chapter (#1298)

ariordan-redhat:
Backports #1298 to 2.4
Adds 2 chapters to the devtools doc.


g-murray:
LGTM

PR #1302, opened 2024-05-01
including proc in the assembling

g-murray:
Linking the set up procedure to the install assembly. 

Location aap-clouds-latest > Stories > assembly-saas-install.adoc


sayjadha (commented):
LGTM.

PR #1308, opened 2024-05-02
Add attributes for Ansible Lightspeed

ariordan-redhat:
Adding Lightspeed attributes to the AAP docs because the DevTools doc refers to Lightspeed.


sayjadha (commented):
LGTM.

PR #1309, opened 2024-05-02
2.4 Add attributes for Ansible Lightspeed (#1308)

ariordan-redhat:
Backports #1308 to 2.4

Adds Lightspeed attributes to the platform docs because the DevTools doc refers to the Lightspeed feature.


sayjadha (commented):
LGTM.

PR #1313, opened 2024-05-03
Corrected Attribute Name (#1312)

KylaF8:
The attribute given doesn't exist and was replaced with {ExecEnvShort}s

AAP-23620

https://issues.redhat.com/browse/AAP-23620


hherbly (approved):
LGTM!

PR #1318, opened 2024-05-03
Remove snippets for tertiary menus

dcdacosta:
This PR deletes the snippets for tertiary menus because the new docs experience allows them to be coded with the menu:selection1[selection2 > selection3] and makes the snippets an unnecessary workaround.


hherbly (approved):
LGTM!

PR #1328, opened 2024-05-07
Removed duplicate attribute definition

dcdacosta:
This PR fixes an issue with the same attribute used more than once. The duplicate attributes (MenuAMCredentials and MenuAMCredentialType) have been removed and defined only once in the file.

### DIFF
--- a/downstream/attributes/attributes.adoc
+++ b/downstream/attributes/attributes.adoc
@@ -163,7 +163,7 @@
 :MenuAEAdminWorkflowApprovals: menu:Views[Workflow Approvals]
 :MenuAETemplates: menu:Resources[Templates]
 :MenuAEProjects: menu:Resources[Projects]
-:MenuInfrastructureInventories: menu:Resoures[Inventories]
+:MenuInfrastructureInventories: menu:Resources[Inventories]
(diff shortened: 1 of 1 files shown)


downstream/attributes/attributes.adoc
```diff
@@ -162,15 +162,13 @@
 :MenuAEAdminActivityStream: menu:Views[Activity Stream]
 :MenuAEAdminWorkflowApprovals: menu:Views[Workflow Approvals]
 :MenuAETemplates: menu:Resources[Templates]
-:MenuAMCredentials: menu:Resources[Credentials]
 :MenuAEProjects: menu:Resources[Projects]
 :MenuInfrastructureInventories: menu:Resoures[Inventories]
```
EMcWhinn, line 166: Hey @dcdacosta, I noticed this attribute has "Resources" spelled wrong.

EMcWhinn (commented):
One minor typo, otherwise LGTM

downstream/attributes/attributes.adoc
dcdacosta: @EMcWhinn that should be ok because it's now been deleted.

EMcWhinn (approved):
LGTM

PR #1334, opened 2024-05-09
Update Controller UG with 2.5 attributes

ianf77:
Correcting procedures to match 2.5 UI

Controller User Guide UI updates chapters 24-33

https://issues.redhat.com/browse/AAP-23674


downstream/modules/platform/proc-controller-create-notification-template.adoc
```diff
@@ -7,7 +7,7 @@ Use the following procedure to create a notification template.
 .Procedure
 
 . From the navigation panel, select {MenuAEAdminJobNotifications}.
```
EMcWhinn, line 9: I can't remember if we spoke about this, its in the section above too. The UI shows Administration > Notifiers, not Job Notifications, so we probably need to change the attribute.

downstream/modules/platform/ref-controller-isolation-functionality.adoc
```diff
@@ -16,8 +16,7 @@ This can be changed in the *Job Execution Path* field of the *Jobs settings* pag
 AWX_ISOLATION_BASE_PATH = "/opt/tmp"
 ----
 
-If there are any additional directories that should specifically be exposed from the host to the container that playbooks run in, you can
-specify those in the *Paths to Expose to Isolated Jobs* field of the *Jobs settings* page, or in the REST API at `/api/v2/settings/jobs`:
+If there are any additional directories that should specifically be exposed from the host to the container that playbooks run in, you can specify those in the *Paths to Expose to Isolated Jobs* field of the *Jobs settings* page, or in the REST API at `/api/v2/settings/jobs`:
```
EMcWhinn, line 19:
```suggestion
If there are any additional directories that should specifically be exposed from the host to the container that playbooks run in, you can specify those in the *Paths to expose to isolated jobs* field of the *Jobs Settings* page, or in the REST API at `/api/v2/settings/jobs`:
```

EMcWhinn (commented):
A couple of very minor comments otherwise looks good.

downstream/modules/platform/proc-controller-create-notification-template.adoc
ianf77: I talked to Donna about this. I thought she'd fixed it.

ianf77: I'll make the change.

EMcWhinn (approved):
LGTM

PR #1341, opened 2024-05-14
Add CPU and RAM requirements for Postgres in planning guide

michellemacrh:
[Docs] Planning Guide should have the RAM and CPU requirement for Postgres system.

https://issues.redhat.com/browse/AAP-20172


ianf77 (approved):
LGTM

PR #1343, opened 2024-05-15
Plus symbols not rendering correctly in asciidoc

EMcWhinn:
Automation Controller API Overview typo

https://issues.redhat.com/browse/AAP-23917

Affects `titles/controller-api-overview`

### DIFF
--- a/downstream/modules/platform/ref-controller-api-config-settings.adoc
+++ b/downstream/modules/platform/ref-controller-api-config-settings.adoc
@@ -51,7 +51,7 @@ To compose a named URL for this specific resource object by using `NAMED_URL_FOR
 
 * The first part of the URL format is `<name>`, which indicates that you can find the label resource detail in `/api/v2/labels/5/`, and look for `name` field in returned JSON. 
 If you have the `name` field with value `Foo`, then the first part of the unique identifier is `Foo`.
-* The second part of the format is double pluses: {plus}{plus}. 
+* The second part of the format is double plus sign {plus}{plus}. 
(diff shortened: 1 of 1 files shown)


downstream/modules/platform/ref-controller-api-config-settings.adoc
```diff
@@ -51,7 +51,7 @@ To compose a named URL for this specific resource object by using `NAMED_URL_FOR
 
 * The first part of the URL format is `<name>`, which indicates that you can find the label resource detail in `/api/v2/labels/5/`, and look for `name` field in returned JSON. 
 If you have the `name` field with value `Foo`, then the first part of the unique identifier is `Foo`.
-* The second part of the format is double pluses "++". 
+* The second part of the format is double pluses: {plus}{plus}.
```
sayjadha, line 54: Suggest 'double pluses' -> 'double plus sign' because #1: pluses is anyways plural and you're covering it when you mention 'double', and #2: IBM style guide contains a few usages of plus sign with symbol in parenthesis.

sayjadha, line 54: Curious as to why we're using this format as `double pluses: {plus}{plus}`, and the rest of the URL elements are in block formatting. Can we have something like 'The second part of the format is double plus sign `++` for consistency? Please check if I'm changing the accuracy of the content, and incorporate this suggestion only if you think it works here.

sayjadha (commented):
A minor comment, rest LGTM.

sayjadha (commented):
LGTM.

PR #1344, opened 2024-05-15
Update BaseURL attribute to docs.redhat.com site

ariordan-redhat:
Update the {BaseURL} attribute so that links to Red Hat docs point to docs.redhat.com instead of access.redhat.com


EMcWhinn (approved):
LGTM

michellemacrh:
I built the release notes and the planning guide, and verified the links work ok 👍

ariordan-redhat:
Backported to `2.4` because access.redhat.com is no longer working for some users in certain browsers.

PR #1354, opened 2024-05-16
AAP-23926 Remove content from EDA GS guide that points customers to c…

jself-sudoku:
Work that resolves [AAP-23926](https://issues.redhat.com/browse/AAP-23926), per SME: 

- Remove content that points customer to the EDA deployment using an internal database since it no longer exists and is not widely supported. Only point to Chapter 3 in the Installation Guide.
- Also, remove links to two other scenarios that have been removed from the installation guide, per direction from support and products (guidance from Mike Silmser).


jself-sudoku:
Thanks, @msilmser.

hherbly (approved):
LGTM

PR #1359, opened 2024-05-17
AAP-24195: Added info. about languages supported for Ansible Lightspeed prompts

sayjadha:
This PR addresses JIRA https://issues.redhat.com/browse/AAP-24195. 

Changes made:

- Added info. that Lightspeed supports user prompts in English language only, along with a corner case.

### DIFF
--- a/lightspeed/modules/con-lightspeed-about.adoc
+++ b/lightspeed/modules/con-lightspeed-about.adoc
@@ -22,7 +22,7 @@ To use {LightspeedFullName}, your organization must have:
 
 * *Reduces the onboarding learning period for Ansible developers*
 +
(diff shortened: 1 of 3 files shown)


lightspeed/modules/con-lightspeed-about.adoc
rogrange, line 1: I would suggest: "can use English language prompts"

sayjadha: @rogrange : I would tend to agree, but the general PM feedback was to use 'natural language prompts' term to denote that Lightspeed accepts non-YAML or non-coding language prompts. Therefore, I'm hesitant to remove the term out.

rogrange: I thought that might be the case, but I wanted to check with you, hence the comment. I will merge the PR.

PR #1370, opened 2024-05-22
Formatting issue in Controller UG Glossary section

KylaF8:
This should show up as a bold term but is displaying with 3 = like this: === Source Credential

AAP-24465

https://issues.redhat.com/browse/AAP-24465

### DIFF
--- a/downstream/assemblies/platform/assembly-controller-glossary.adoc
+++ b/downstream/assemblies/platform/assembly-controller-glossary.adoc
@@ -231,7 +231,7 @@ The calendar of dates and times for which a job should run automatically.
 See *Distributed Job*.
 
 [discrete]
-*Source Credential*::
+=== Source Credential
 An external credential that is linked to the field of a target credential.
 
 [discrete]
--- a/downstream/assemblies/platform/assembly-controller-glossary.adoc
+++ b/downstream/assemblies/platform/assembly-controller-glossary.adoc
@@ -231,7 +231,7 @@ The calendar of dates and times for which a job should run automatically.
 See *Distributed Job*.
 
 [discrete]
-=== Source Credential::
+*Source Credential*::
 An external credential that is linked to the field of a target credential.
 
 [discrete]


downstream/assemblies/platform/assembly-controller-glossary.adoc
```diff
@@ -231,7 +231,7 @@ The calendar of dates and times for which a job should run automatically.
 See *Distributed Job*.
 
 [discrete]
-=== Source Credential::
+*Source Credential*::
```
michellemacrh, line 234: I think the reason the formatting is broken is because of the two `::`. I wonder if it would make sense to match the formatting of the other sections and have it be `=== Source Credential` instead?

KylaF8: Oh I see, in the ticket it asked for it to be in bold, or that was my understanding of it, I will attach it below: https://issues.redhat.com/browse/AAP-24465

michellemacrh: Ah, it might be worth checking with Donna to see if that is the intended result?

KylaF8: Okay perfect, thank you :)

dcdacosta: Hey @KylaF8, @michellemacrh is correct that you just need to remove the 2 colons at the end. This should be in the file as follows:

=== Source Credential

KylaF8: Hi Donna, okay perfect thank you. So instead of having it as: *Source Credential*:: 
Have it as: ===Source Credential

dcdacosta: @KylaF8 Yes, but don't forget the space between === and Source Credential. So, 

=== Source Credential

annibond (approved):
looks good

annibond (approved):
looks good

dcdacosta (approved):
LGTM!

PR #1371, opened 2024-05-22
AAP-17190 Modify devtools intro and scaffolding

ariordan-redhat:
AAP-17190 Modify devtools intro and scaffolding
Affects `titles/develop-automation-content/`


g-murray:
LGTM

PR #1378, opened 2024-05-23
Formatting issue in Controller UG Glossary section (#1370)

KylaF8:
* Formatting issue in Controller UG Glossary section

This should show up as a bold term but is displaying with 3 = like this: === Source Credential

AAP-24465

https://issues.redhat.com/browse/AAP-24465

* Formatting issue in Controller UG Glossary section

This should show up as a bold term but is displaying with 3 = like this: === Source Credential

AAP-24465

https://issues.redhat.com/browse/AAP-24465


dcdacosta (approved):
LGTM!

PR #1383, opened 2024-05-29
AAP-24690 Tweaked note after step 5 in Chapter 5 EDA controller user guide

jself-sudoku:
For the EDA controller user guide, chapter 5.1, tweaked the note after step 5 to include permissions to execute job and workflow templates. (Note: this update will occur in the 2.4 guide, but will be applied to 2.5 separately since the process will change slightly.)

### DIFF
--- a/downstream/modules/eda/proc-eda-set-up-token.adoc
+++ b/downstream/modules/eda/proc-eda-set-up-token.adoc
@@ -24,7 +24,7 @@ For more information about creating the token, see the link:https://docs.ansible
 +
 [NOTE]
 ====
-The token you create must have write permissions, and the assigned user must have permission to execute job and workflow templates.
+Your token must be in write-scope and if you are the assigned user, you must have permission to execute job and workflow templates.
 ====
 . Select btn:[Create controller token].
 
--- a/downstream/modules/eda/proc-eda-set-up-token.adoc
+++ b/downstream/modules/eda/proc-eda-set-up-token.adoc
@@ -1,6 +1,6 @@
 [id="eda-set-up-token-to-authenticate"]
 
-= Setting up a token to authenticate to {PlatformNameShort} Controller
+= Setting up a token to authenticate to {ControllerName}
 
 .Prerequisites
 
--- a/downstream/modules/eda/proc-eda-set-up-token.adoc
+++ b/downstream/modules/eda/proc-eda-set-up-token.adoc
@@ -24,7 +24,7 @@ For more information about creating the token, see the link:https://docs.ansible
 +
 [NOTE]
 ====
-The token must be in write-scope.
+The token you create must have write permissions, and the assigned user must have permission to execute job and workflow templates.
 ====
 . Select btn:[Create controller token].
 


downstream/modules/eda/proc-eda-set-up-token.adoc
```diff
@@ -24,7 +24,7 @@ For more information about creating the token, see the link:https://docs.ansible
 +
 [NOTE]
 ====
-The token must be in write-scope.
+The token you create must have write permissions, and the assigned user must have permission to execute job and workflow templates.
```
Alex-Izquierdo, line 27: "job template" is a different entity than "workflow template". Is maybe better to not omit the word "template" to avoid misleading? 

```suggestion
The token you create must have write permissions, and the assigned user must have permission to execute job templates and workflow templates.
```

mkanoor: @Alex-Izquierdo Can we get a screen shot of the minimal permission from the AAP UI so we don't have any ambiguities.

Alex-Izquierdo: I think that is a call for @jself-sudoku 
In the existing documentation we don't have specific instructions about controller, we just point to a link of their documentation. 

If you want to add a screenshot I can take it for you.

jself-sudoku: @Alex-Izquierdo - I didn't delete a word there. I combined job templates and workflow templates, which is a grammatical and style rule. Instead of saying template twice, we just say job and workflow templates. 

@mkanoor - all docs teams are encouraged to use as few screenshots as possible due to the work it takes to maintain them, specifically if screen shots change. Can you let me know what ambiguities you're referring to in this instance so we can minimize that?

Alex-Izquierdo (approved):

downstream/modules/eda/proc-eda-set-up-token.adoc
jself-sudoku: @Alex-Izquierdo and @mkanoor - Just an FYI. The docs team request SME reviews for technical accuracy. Editorial and peer reviews focus more on grammar and style. We usually handle those after we ensure that the content is technically accurate. Can one or both of you provide approval so that we can move this quick change to the next phase of our processes and then ensure that the change is also applied for 2.5?

mkanoor: @Alex-Izquierdo @jself-sudoku I don't know how to create tokens, there is an older version of the UI and a newer version. I tried looking in the new UI and didn't see an option under Users to create a Token. Is this different in the newer UI?

jself-sudoku: Hi @mkanoor. I ran into the same issue, which is why I asked the question about it in the Slack discussion [here](https://redhat-internal.slack.com/archives/C04930GDCTS/p1716426880812629). You can see my question [here](https://redhat-internal.slack.com/archives/C04930GDCTS/p1716931145237329?thread_ts=1716426880.812629&cid=C04930GDCTS), which is now crossed out since my teammate, @dcdacosta , shared a link to the latest [mockups]([https://www.figma.com/proto/gMrHPRqYwvZw5yrkL8tXHF/Access-Management?page-id=82%3A8[…]port=1257%2C-3512%2C0.61&t=Yzn8joBu2pzCzvui-1&scaling=min-zoom](https://www.figma.com/proto/gMrHPRqYwvZw5yrkL8tXHF/Access-Management?page-id=82%3A89874&type=design&node-id=1690-82285&viewport=1257%2C-3512%2C0.61&t=Yzn8joBu2pzCzvui-1&scaling=min-zoom)). There you can see what is being planned to carry out this process for 2.5. The tab is simply referred to as **Tokens** in the mockups.

Alex-Izquierdo (approved):
I like more the previous version, but style matters are not my call. Technically is correct.

downstream/modules/eda/proc-eda-set-up-token.adoc
```diff
@@ -24,7 +24,7 @@ For more information about creating the token, see the link:https://docs.ansible
 +
 [NOTE]
 ====
-The token must be in write-scope.
+Your token must be in write-scope and if you are the assigned user, you must have permission to execute job and workflow templates.
 ====
 . Select btn:[Create controller token].
```
dcdacosta, line 29: There should be a .Procedure before the steps.

jself-sudoku: Hi @dcdacosta. This peer review was or the note only. There's already a .Procedure before the steps on line 11 above if you expand the section. Is that want you mean?

downstream/modules/eda/proc-eda-set-up-token.adoc
```diff
@@ -24,7 +24,7 @@ For more information about creating the token, see the link:https://docs.ansible
 +
 [NOTE]
 ====
-The token must be in write-scope.
+Your token must be in write-scope and if you are the assigned user, you must have permission to execute job and workflow templates.
 ====
```
dcdacosta, line 28: I think you need a + after the closing ==== for the note to make sure the numbering is continued but I'm not %100 certain.

dcdacosta (approved):
One minor change, otherwise, LGTM!

downstream/modules/eda/proc-eda-set-up-token.adoc
dcdacosta: Ignore this. I built it locally and it formats ok. I'll proceed with merging.

dcdacosta (approved):

jself-sudoku:
Note to anyone checking builds: this merge was reverted as noted above in PR #1771. It was an old PR that was put on hold due to chapter on controller token being deprecated due to confusion for the customer. Later, we were advised that we could still add this note to the 2.4 EDA controller doc. We proceeded, but realized that the `main` branch we're now using is 2.5. So we reverted the update here and applied the update directly to the upstream/2.4 branch. No need to add backport labels here.

PR #1385, opened 2024-05-30
AAP-24694 Added note about rulebook activations shutting down and canceling jobs on controller.

jself-sudoku:
Added a note at the end of [section 6.1 Setting up a rulebook activation](https://access.redhat.com/documentation/en-us/red_hat_ansible_automation_platform/2.4/html-single/event-driven_ansible_controller_user_guide/index#eda-set-up-rulebook-activation) in the EDA controller user guide to inform users that when an activation shuts down, any tasks that are waiting to be performed will be canceled and an info level message will be sent to the activation log. Tracked in jira [AAP-24694](https://issues.redhat.com/browse/AAP-24694).


benthomasson (changes requested):
I would remove "This is expected behavior." as this is implied by inclusion of the previous statements in the documentation.

benthomasson (approved):

michellemacrh:
LGTM!

PR #1415, opened 2024-06-11
Added note to say usability analytics isn't supported

ianf77:
Remove User Analytics chapter from Controller Admin Guide

https://issues.redhat.com/browse/AAP-25200


michellemacrh:
LGTM

PR #1420, opened 2024-06-11
Clouds: Update BaseURL attribute

ariordan-redhat:
Updates `BaseURL` attribute for the Clouds branch.
Fixes a link in Azure docs that uses the attribute.


g-murray:
LGTM

PR #1424, opened 2024-06-12
Removing AAP acronyms

g-murray:
Removing the incorrect use of AAP acroynms from "ref-network-ports-protocols.adoc"


michellemacrh:
LGTM!

PR #1425, opened 2024-06-12
2.4 Add note about collection precedence when running a job (#1423)

michellemacrh:
Update docs to list order of precedence if project is using execution environment and requirements.yml file

https://issues.redhat.com/browse/AAP-20646


g-murray:
LGTM

PR #1426, opened 2024-06-12
Add boilerplate module

ariordan-redhat:
Create a boilerplate module that contains the front matter for all AAP docs,
so that 
we can edit this content in one place and have it replicated in all docs.


michellemacrh:
LGTM!

PR #1441, opened 2024-06-14
AAP-25229 Update URL for link from Azure doc to mesh doc

ariordan-redhat:
AAP-25229 Update link from Azure doc to mesh doc


ianf77 (approved):
LGTM

PR #1461, opened 2024-06-18
AAP-20958 Lightspeed Set grandparent-context for assembly with nesting

ariordan-redhat:
AAP-20958 Lightspeed Set `grandparent-context` value for assembly that contains nested assemblies.

When an assembly contains nested assemblies, the `parent-context` variable is set by the nested assemblies.
After the nested assemblies, you need to restore the context to the original `parent-context` (ie the value before the nesting).
This PR does this by creating a variable called `grandparent-context`.


sayjadha:
This is new learning for me so thank you @ariordan-redhat for the knowledge. Merging the PR.

PR #1463, opened 2024-06-20
PAA-25650 Add title for AAP plugins for RHDH

ariordan-redhat:
Add a new title for AAP plugins for RHDH.
Scaffold assemblies with placeholder content, so that we can avoid merge conflicts when simultaneously working on separate PRs for each assembly.


g-murray:
LGTM

PR #1474, opened 2024-06-24
Setting up new proc files for AAP-22178

g-murray:
Getting a start on the backend of AAP-22178 while the tech review is on going

Adding new procs to:

- assembly-aap-backup.adoc
- assembly-aap-recovery.adoc


downstream/modules/platform/proc-aap-platform-gateway-restore.adoc
```diff
@@ -0,0 +1,3 @@
+[id="aap-platform-gateway-restore"]
+
+= Recovering a Red Hat Ansible Automation Platform deployment
```
ianf77, line 3: You could probably use {PlatformName} here.

ianf77 (commented):
One minor suggestion. Can be ignored

PR #1479, opened 2024-06-24
AAP-25788: Incorporate SME feedback and language edits for 'Configuring custom models' chapter

sayjadha:
This PR addresses JIRA https://issues.redhat.com/browse/AAP-25788. 

Changes made:

- Incorporated SME feedback
- Language edits found during self review

### DIFF
--- a/lightspeed/assemblies/assembly_configuring-custom-models.adoc
+++ b/lightspeed/assemblies/assembly_configuring-custom-models.adoc
@@ -12,13 +12,22 @@ ifdef::context[:parent-context: {context}]
 
 As an organization administrator, you can create and use fine-tuned, custom models that are trained on your organization's existing Ansible content. With this capability, you can tune the models to your organization's automation patterns and improve the code recommendation experience.
 
-You can configure multiple custom models for your organization. For example, you can create a custom model for your corporate IT automation team and a different one for your engineering team's infrastructure. You can also configure a custom model to make it available for all Ansible users or select Ansible users in your organization. 
+After you create a custom model, you can specify one of the following access types:
+
+* Enable access for all users in your organization
++
+You can configure the custom model as the default model for your organization. All users in your organization can use the custom model.
+
+* Enable access for select Ansible users in your organization
++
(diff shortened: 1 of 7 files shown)


lightspeed/assemblies/assembly_configuring-custom-models.adoc
rogrange, line 1: Not certain how big a deal this is, however, for "Using the model-override setting in the Ansible VS Code extension, select Ansible users only can override the default model setting and tune their Ansible Lightspeed service to use the custom model."

I would place a full-stop after 'Code extension' and rephrase the next sentence to read "Only select Ansible users can override. . . '

It is correct the way it stands, I thought the rephrasing might be clearer for the user.

lightspeed/modules/proc_configure-custom-models-lightspeed.adoc
rogrange, line 1: Same comment here as above.

lightspeed/assemblies/assembly_configuring-custom-models.adoc
sayjadha: @rogrange : "Using the model-override setting in the Ansible VS Code extension" doesn't sound like a complete sentence. Offering the following suggestions, which one would you recommend?

#1: Using the model-override setting in the Ansible VS Code extension, select Ansible users can override the default model and tune their Ansible Lightspeed service to use the custom model instead.

#2: Using the model-override setting in the Ansible VS Code extension, select Ansible users can tune their Ansible Lightspeed service to use the custom model instead of the default model.

rogrange: I would go with #2.

sayjadha: This is done, @rogrange. Please review?

lightspeed/modules/proc_configure-custom-models-lightspeed.adoc
sayjadha: Done.

PR #1485, opened 2024-06-25
AAP-25806 Azure - update CIDR range minimums

ariordan-redhat:
AAP-25806 Azure - update CIDR range minimums for Service CIDR and Pod CIDR
Affects `/titles/aap-on-azure`


dcdacosta (approved):
LGTM!

PR #1496, opened 2024-06-27
AAP-25889 Azure - link to kcs article for list of domains

ariordan-redhat:
AAP-25889 Azure - link to a KCS article that customers / support can refer to for the list of domains to allow, instead of listing the domains in the docs, because this list is frequently updated.


sayjadha (commented):
LGTM.

PR #1504, opened 2024-06-28
Commit for AAP-22613 Remove SSO deployment references

rogrange:
This commit encompasses AAP-22613, it also includes removing any reference to LDAP configuration. Both SSO and LDAP refernces are being removed due to an approach change to installation of the AAP platform.

Resolves: AAP-22613


simaishi:
Also need to remove section about enabling `jb-eap-7.3-for-rhel-8-x86_64-rpms` and `rh-sso-7.4-for-rhel-8-x86_64-rpms` repos from proc-synchronizing-rpm-repositories-by-using-reposync.adoc, but the rest looks good for installer.  Searching in this repo, there are still a few mention of SSO outside of the installer guide, so that might need to be taken care of separately.

downstream/modules/platform/ref-controller-variables.adoc
```diff
@@ -8,7 +8,10 @@
 | *`admin_password`* | The password for an administration user to access the UI when the installation is complete.
 
 Passwords must be enclosed in quotes when they are provided in plain text in the inventory file.
-| *`automation_controller_main_url`* | For an alternative front end URL needed for SSO configuration, provide the URL.
+
+| *`automation_controller_main_url`* | The full routeable URL used by EDA to connect to a controller host. This URL is required if there is no Automation Controller configured in inventory.
```
sayjadha, line 12: 'Automation Controller' -> 'automation controller'. I did a quick scan of the usage throughout the guide and sentence case is used in most instances.

sayjadha (commented):
@rogrange : One minor suggestion, rest LGTM.

downstream/modules/platform/ref-controller-variables.adoc
rogrange: in this particular instance,  the noted content is a variable and does not use sentence case.

sayjadha:
@rogrange mentioned that in this particular instance,  the noted content is a variable and does not use sentence case. No update needed so I'm approving this PR.

simaishi:
This is for 2.5 only, no backport required.

PR #1518, opened 2024-07-02
2.4 AAP-25848 Add Microsoft Entra ID attribute to downstream docs (#1517)

ariordan-redhat:
AAP-25848 Add an attribute for Microsoft Entra ID to the downstream docs.


ianf77 (approved):
LGTM

PR #1524, opened 2024-07-03
Commit for AAP-22613 Remove SSO deployment references (#1504)

rogrange:
This commit encompasses AAP-22613, it also includes removing any reference to LDAP configuration. Both SSO and LDAP refernces are being removed due to an approach change to installation of the AAP platform.

Resolves: AAP-22613


downstream/modules/platform/ref-controller-variables.adoc
```diff
@@ -8,7 +8,10 @@
 | *`admin_password`* | The password for an administration user to access the UI when the installation is complete.
 
 Passwords must be enclosed in quotes when they are provided in plain text in the inventory file.
-| *`automation_controller_main_url`* | For an alternative front end URL needed for SSO configuration, provide the URL.
+
+| *`automation_controller_main_url`* | The full routeable URL used by EDA to connect to a controller host. This URL is required if there is no Automation Controller configured in inventory.
```
dcdacosta, line 12: Automation Controller should use the attribute {ControllerName}. Also, automation controller is a component and not a product so it is not a proper noun. This change is minor so I can merge anyway if you want to fix in a separate PR?

dcdacosta (commented):
Minor nit, let me know if you want me to proceed with a merge.

downstream/modules/platform/ref-controller-variables.adoc
rogrange: This is a variable, and not a direct reference to automation controller. As far as I understand it, variables are not subject to sentence case or attributes. If I am wrong, I will fix this in another PR.

dcdacosta (approved):

PR #1526, opened 2024-07-03
AAP-23395: Reused  known issue about using remote SSH extension from existing inline troubleshooting doc

sayjadha:
This PR addresses AAP-23395, which was originally added to ongoing troubleshooting inline messages (AAP-22335) effort, but reusing it here because the troubleshooting activity is still WIP and delayed than originally planned. Content was reviewed as part of the ongoing error messaging effort. 

Change made:
Added error message about Ansible Lightspeed Task Generation in Playbooks does not work with Remote SSH Extension for workspaces with large number of roles.


dcdacosta (approved):
LGTM!

PR #1528, opened 2024-07-03
Fixed naming content with proper attribute. (#1527)

rogrange:
Replaced Automation controller with {ControllerName}, and EDA with {EDAName}.


dcdacosta (approved):
LGTM!

PR #1538, opened 2024-07-05
Adds new EDA 2.5 w/controller 2.4 adoc. Updates main assembly file.

rogrange:
Adding new file and content for installing EDA 2.5 with controller 2.4. Updated the assembly-platform-install-scenario.adoc to reflect the new file.

Resolves: AAP-23518

### DIFF
--- a/downstream/modules/platform/con-eda-2-5-with-controller-2-4.adoc
+++ b/downstream/modules/platform/con-eda-2-5-with-controller-2-4.adoc
@@ -6,7 +6,7 @@
 [id="eda-2-5-with-controller-2-4_{context}"]
 = {EDAName} 2.5 with controller 2.4
 
-Use the following example to populate the inventory file to deploy a new single instance of {EDAName} 2.5 with controller 2.4. For {EDAName}, the requirements for a connection to controller are the `automation_controller_main_url` pointing to the 2.4 controller URL and `username/password`.
+Use the following example to populate the inventory file to deploy a new single instance of {EDAName} 2.5 with controller 2.4. For {EDAName}, the requirements for a connection to controller are the `automation_controller_main_url` pointing to the 2.4 controller URL.
 
 
 ----
--- a/downstream/modules/platform/con-eda-2-5-with-controller-2-4.adoc
+++ b/downstream/modules/platform/con-eda-2-5-with-controller-2-4.adoc
@@ -11,8 +11,10 @@ Use the following example to populate the inventory file to deploy a new single 
 
 ----
 [automationedacontroller]
+eda.example.org
 
 [automationgateway]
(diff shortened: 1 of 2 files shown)


downstream/modules/platform/con-eda-2-5-with-controller-2-4.adoc
```diff
@@ -0,0 +1,59 @@
...
+
+[id="eda-2-5-with-controller-2-4_{context}"]
+= {EDAName} 2.5 with controller 2.4
+
+Use the following example to populate the inventory file to deploy a new single instance of {EDAName} 2.5 with controller 2.4. For {EDAName}, the requirements for a connection to controller are the `automation_controller_main_url` pointing to the 2.4 controller URL and `username/password`.
+
+
+----
+[automationedacontroller]
+
```
jamesmarshall24, line 14: We need an example hostname here where the installer will install EDA

downstream/modules/platform/con-eda-2-5-with-controller-2-4.adoc
```diff
@@ -0,0 +1,59 @@
...
+[automationgateway]
+
```
jamesmarshall24, line 16: We also need a hostname here, but since we're doing a single EDA installation, the hostname you use for line 14 should be the same host here on line 16. That way EDA and gateway will install on the same host.

downstream/modules/platform/con-eda-2-5-with-controller-2-4.adoc
```diff
@@ -0,0 +1,59 @@
+:_newdoc-version: 2.18.3
+:_template-generated: 2024-07-05
+
+:_mod-docs-content-type: CONCEPT
+
+[id="eda-2-5-with-controller-2-4_{context}"]
+= {EDAName} 2.5 with controller 2.4
+
+Use the following example to populate the inventory file to deploy a new single instance of {EDAName} 2.5 with controller 2.4. For {EDAName}, the requirements for a connection to controller are the `automation_controller_main_url` pointing to the 2.4 controller URL and `username/password`.
```
jamesmarshall24, line 9:
```suggestion
Use the following example to populate the inventory file to deploy a new single instance of {EDAName} 2.5 with controller 2.4. For {EDAName}, the requirements for a connection to controller are the `automation_controller_main_url` pointing to the 2.4 controller URL.
```

If we want to mention authentication credentials for controller, then we should link to: https://docs.ansible.com/automation-controller/latest/html/userguide/users.html#users-tokens otherwise its not needed in the installation inventory file at this time.

jamesmarshall24 (approved):

annibond:
lgtm

PR #1541, opened 2024-07-08
AAP-24489 Add VS Code remote icon to images dir

ariordan-redhat:
Add an image to the images dir for the remote icon that appears in the status bar of VS Code.


hherbly (approved):
LGTM!

PR #1569, opened 2024-07-11
Move admin username and password to gateway docs, consolidate variables

AlanCoding:
These are docs changes corresponding to https://github.com/ansible/automation-platform-collection/pull/1201

I believe the `automationcontroller_password` variable was incorrect and never worked.

Other variables, like `automationedacontroller_admin_username` I am intentionally removing as a part of the linked work. The EDA variables, only, would require a specific changelog entry that we stopped supporting them.

Gateway variables like `automationgateway_admin_password` were never in released product, so those are being deleted by me, but require no changelog.


AlanCoding:
Filing in Jira as separated issues.

PR #1583, opened 2024-07-16
Added note about removal of container images (#1580)

ianf77:
Need to explain container images will be removed when upgrading AAP

https://issues.redhat.com/browse/AAP-25760


michellemacrh:
LGTM

PR #1589, opened 2024-07-16
AAP-27260 Add attributes for docs titles

ariordan-redhat:
Create attributes for docs titles so that external links to other docs in the AAP docset don't break when titles are updated.


michellemacrh:
LGTM - I like the idea of using the directory structure for the title naming and using Title at the start to easily identify the attributes for titles 👍

ariordan-redhat:
Thanks, @michellemacrh , I'll do the rest of the titles.

g-murray (approved):
LGTM

PR #1590, opened 2024-07-16
AAP-27282 - Updated attribute for credentials

dcdacosta:
This PR updates the attributes for credentials and credential types to be distinct for controller and EDA.


hherbly (approved):
LGTM!

PR #1593, opened 2024-07-17
List RHEL 9.0 as supported in planning guide

michellemacrh:
Explicitly list RHEL 9.0 as supported in the base system requirements in the planning guide

Explicitly list RHEL 9 as supported

https://issues.redhat.com/browse/AAP-26227


EMcWhinn (approved):
LGTM

PR #1608, opened 2024-07-18
Added note to API guide for filtering exact matches (#1606)

EMcWhinn:
Addresses API behavior change from previous release regarding default exact matches.

https://issues.redhat.com/browse/AAP-27347

Affects `titles/controller-api-overview`


sayjadha (commented):
LGTM.

PR #1612, opened 2024-07-22
updating feedback and disclaimer modules in aap-common

g-murray:
Updating the feedback and disclaimer modules with aap-common


ariordan-redhat (approved):
LGTM

PR #1614, opened 2024-07-22
Correct Vale errors in our aap-docs

michellemacrh:
Some errors of note that should be resolved as part of this ticket:

- Keep one space between words
- Use 'Ansible Playbook' rather than 'Ansible playbook'.
- Use 'before' rather than 'prior to'.
- Use 'built-in' rather than 'built in'.
- Use 'checkbox' rather than 'check box'.
- Use 'environment' rather than 'ENV'.
- Use 'includes' rather than 'ships with'.
- Use 'log in to' rather than 'Log into'.
- Use 'pop-up' rather than 'popup'.
- Use 'want' rather than 'wish'.

Correct Vale errors in our aap-docs

https://issues.redhat.com/browse/AAP-24651


downstream/modules/platform/proc-installing-the-ansible-builder-rpm.adoc
```diff
@@ -17,11 +17,11 @@ On the RHEL system where custom {ExecEnvShort}s will be built, you will install
 
 .. Subscribe the RHEL system to a Satellite on the disconnected network.
 
-.. Attach the {PlatformNameShort} subscription and enable the AAP repository.  The repository name will either be `ansible-automation-platform-2.4-for-rhel-8-x86_64-rpms` or `ansible-automation-platform-2.4-for-rhel-9-x86_64-rpms` depending on the version of RHEL used on the underlying system.
+.. Attach the {PlatformNameShort} subscription and enable the {PlatformNameShort} repository. The repository name will either be `ansible-automation-platform-2.4-for-rhel-8-x86_64-rpms` or `ansible-automation-platform-2.4-for-rhel-9-x86_64-rpms` depending on the version of RHEL used on the underlying system.
```
ianf77, line 20: will either be -> is either

ianf77 (commented):
Just one comment

ianf77 (approved):

PR #1616, opened 2024-07-22
Correct Vale errors in our aap-docs (#1614)

michellemacrh:
* Correct Vale errors in our aap-docs

Some errors of note that should be resolved as part of this ticket:

- Keep one space between words
- Use 'Ansible Playbook' rather than 'Ansible playbook'.
- Use 'before' rather than 'prior to'.
- Use 'built-in' rather than 'built in'.
- Use 'checkbox' rather than 'check box'.
- Use 'environment' rather than 'ENV'.
- Use 'includes' rather than 'ships with'.
- Use 'log in to' rather than 'Log into'.
- Use 'pop-up' rather than 'popup'.
- Use 'want' rather than 'wish'.

Correct Vale errors in our aap-docs

https://issues.redhat.com/browse/AAP-24651

* updates based on peer review feedback


hherbly (approved):
LGTM!

PR #1625, opened 2024-07-24
updating apap-common disclaimer

g-murray:
Just updating aap-common disclaimer to adhere to the style guide and to be consistent with main and saas docs.


EMcWhinn (approved):
LGTM

PR #1667, opened 2024-08-01
AAP-23515 OpenSSL deficiency

annibond:
[AAP-23515](https://issues.redhat.com/browse/AAP-23515): Added a warning about patch 3.2 for OpenSSL that makes Ansible automation crash upon load. It requires the administrator to change the API call used for every user in the system.

If that double warning is showing up for you, please delete it. Merge & squash if possible. I can't consistently get those two repeating chunks to show.

### DIFF
--- a/downstream/modules/platform/ref-controller-database-settings.adoc
+++ b/downstream/modules/platform/ref-controller-database-settings.adoc
@@ -15,6 +15,13 @@ To improve the performance of the PostgreSQL server, configure the following _Gr
 
 * `shared_buffers`: determines how much memory is dedicated to the server for caching data. The default value for this parameter is 128 MB. When you modify this value, you must set it between 15% and 25% of the machine's total RAM. 
 
+WARNING
+====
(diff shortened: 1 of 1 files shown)


downstream/modules/platform/ref-controller-database-settings.adoc
```diff
@@ -15,7 +15,24 @@ To improve the performance of the PostgreSQL server, configure the following _Gr
...
 
-NOTE: You must restart the database server after changing the value for shared_buffers.
+WARNING
+====
+If you are compiling Postgres against OpenSSL 3.2, your system regresses to remove the parameter for User during startup. You can rectify this by using the BIO_get_app_data call instead of open_get_data. Only an administrator can make these changes, but it impacts all users connected to the PostgreSQL database.
+
+If you update your systems without the OpenSSL patch, you are not impacted, and you do not need to take action.
+====
+
+WARNING
```
jself-sudoku, line 25: As you mentioned, there's a duplication of the admonition. Can you see it in VSCode? I think we should remove the dup before we merge, if that's possible.

jself-sudoku: Otherwise, LGTM. I can squash/merge when the correction is made.

jself-sudoku:
LGTM

PR #1692, opened 2024-08-06
2.5 AAP-27260 Add attributes for docs titles (#1589)

ariordan-redhat:
Create attributes for docs titles so that external links to other docs in the AAP docset don't break when titles are updated.
Backports #1589


michellemacrh (approved):
LGTM

PR #1718, opened 2024-08-12
Add admonition to use of installer setup script

ianf77:
Update text in docs to tell users ansible-core 2.16 must be installed before running setup.sh

https://issues.redhat.com/browse/AAP-29119


g-murray (approved):
LGTM

PR #1722, opened 2024-08-12
AAP-21827 Add new field for new field for the k8s_service_name

jself-sudoku:
Just one update made for the Create Rulebook Activations fields: Service Name. Other more extensive changes will be tracked in a separate PR.


g-murray (approved):
LGTM

PR #1735, opened 2024-08-13
Document the Metric Collection Utility

KylaF8:
https://issues.redhat.com/browse/AAP-19692

AAP-19692


hherbly (approved):
This will need some editing, but I'm approving and merging to add it to the repo. Further edits can be made after that.

PR #1747, opened 2024-08-15
Document the requirement that upgrade from 2.4 containerized TP is not supported

michellemacrh:
Document the requirement that upgrade from 2.4 containerized TP is not supported

Created a snippet for this text /snippets/container-upgrades.adoc

https://issues.redhat.com/browse/AAP-28343


ianf77 (approved):
LGTM

PR #1750, opened 2024-08-15
2.4 AAP-29354 Add Tech Preview notice to Devtools doc

ariordan-redhat:
Adding tech preview boilerplate notice for devtools doc to to 2.4 only, as devtools will be GA in 2.5.
Affects `/titles/develop-automation-content`


michellemacrh:
LGTM!

PR #1772, opened 2024-08-21
AAP-24690 Tweaked note about write-scope status for controller token

jself-sudoku:
For the EDA controller user guide, chapter 5.1, tweaked the note after step 5 to include permissions to execute job and workflow templates. (Note: this update will occur in the 2.4 guide, but will be applied to 2.5 separately since the process will change slightly.)


dcdacosta (approved):
LGTM!

PR #1780, opened 2024-08-22
Added warning to "About inventory installer file" AAP-15722

annibond:
Added warning that the air-gapped exception to the deprecation of manifests, which currently applies to all other use cases, is disappearing with Satellite 6.16 as per AAP-15722.

PR #1792, opened 2024-08-28
Backport of AAP-15722: about inventory install file

annibond:
Backport of PR #1780

Added warning that the air-gapped exception to the deprecation of manifests, which currently applies to all other use cases, is disappearing with Satellite 6.16 as per AAP-15722.

### DIFF
--- a/downstream/modules/platform/ref-controller-database-settings.adoc
+++ b/downstream/modules/platform/ref-controller-database-settings.adoc
@@ -22,7 +22,7 @@ You must restart the database server after changing the value for `shared_buffer
 
 [NOTE]
 ====
-If you are compiling Postgres against OpenSSL 3.2, your system regresses to remove the parameter for User during startup. You can rectify this by using the BIO_get_app_data call instead of open_get_data. Only an administrator can make these changes, but it impacts all users connected to the PostgreSQL database.
+If you are compiling Postgres against OpenSSL 3.2, your system regresses to remove the parameter for User during startup. You can rectify this by using the `BIO_get_app_data` call instead of `open_get_data`. Only an administrator can make these changes, but it impacts all users connected to the PostgreSQL database.
 
 If you update your systems without the OpenSSL patch, you are not impacted, and you do not need to take action.
 ====
--- a/downstream/modules/platform/ref-controller-database-settings.adoc
+++ b/downstream/modules/platform/ref-controller-database-settings.adoc
@@ -25,6 +25,13 @@ If you are compiling Postgres against OpenSSL 3.2, your system regresses to remo
 You must restart the database server after changing the value for `shared_buffers`.
 ====
 
+Warning
+====
+If you are compiling Postgres against OpenSSL 3.2, your system regresses to remove the parameter for User during startup. You can rectify this by using the BIO_get_app_data call instead of open_get_data. Only an administrator can make these changes, but it impacts all users connected to the PostgreSQL database.
+
+If you update your systems without the OpenSSL patch, you are not impacted, and you do not need to take action.
+====
+
 * `work_mem`: provides the amount of memory to be used by internal sort operations and hash tables before disk-swapping. Sort operations are used for order by, distinct, and merge join operations. Hash tables are used in hash joins and hash-based aggregation. The default value for this parameter is 4 MB. Setting the correct value of the `work_mem` parameter improves the speed of a search by reducing disk-swapping.
 ** Use the following formula to calculate the optimal value of the `work_mem` parameter for the database server: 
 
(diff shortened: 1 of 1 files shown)


downstream/modules/platform/ref-controller-database-settings.adoc
```diff
@@ -25,6 +25,13 @@ If you are compiling Postgres against OpenSSL 3.2, your system regresses to remo
 You must restart the database server after changing the value for `shared_buffers`.
 ====
 
+Warning
```
ianf77, line 28: I think this should be [WARNING]

downstream/modules/platform/ref-controller-database-settings.adoc
```diff
@@ -25,6 +25,13 @@ If you are compiling Postgres against OpenSSL 3.2, your system regresses to remo
...
+====
+If you are compiling Postgres against OpenSSL 3.2, your system regresses to remove the parameter for User during startup. You can rectify this by using the BIO_get_app_data call instead of open_get_data. Only an administrator can make these changes, but it impacts all users connected to the PostgreSQL database.
```
ianf77, line 30: for the user

downstream/modules/platform/ref-controller-database-settings.adoc
annibond: This was supposed to be a note actually. Jameria and I had a longer conversation about it and it wasn't catastrophic to the user, so we landed on NOTE, so I'm not sure why it didn't follow this copy.

downstream/modules/platform/ref-controller-database-settings.adoc
annibond: I have made the changes for both BIO_get_app_data and the second API call in the same paragraph. Thank you!

downstream/modules/platform/ref-controller-database-settings.adoc
```diff
@@ -17,14 +17,21 @@ To improve the performance of the PostgreSQL server, configure the following _Gr
 
 [WARNING]
 ====
-If you are compiling Postgres against OpenSSL 3.2, your system regresses to remove the parameter for User during startup. You can rectify this by using the BIO_get_app_data call instead of open_get_data. Only an administrator can make these changes, but it impacts all users connected to the PostgreSQL database. f you update your systems without the OpenSSL patch, you are not impacted, and you do not need to take action.
+If you are compiling Postgres against OpenSSL 3.2, your system regresses to remove the parameter for User during startup. You can rectify this by using the `BIO_get_app_data` call instead of `open_get_data`. Only an administrator can make these changes, but it impacts all users connected to the PostgreSQL database. f you update your systems without the OpenSSL patch, you are not impacted, and you do not need to take action.
```
jself-sudoku, line 20:
```suggestion
If you are compiling Postgres against OpenSSL 3.2, your system regresses to remove the parameter for User during startup. You can rectify this by using the `BIO_get_app_data` call instead of `open_get_data`. Only an administrator can make these changes, but it impacts all users connected to the PostgreSQL database. If you update your systems without the OpenSSL patch, you are not impacted, and you do not need to take action.
```

downstream/modules/platform/ref-controller-database-settings.adoc
jself-sudoku: I think this note is what Ian was referring to. I have a question. Is this note a duplication of the Warning above? Should this one be removed?

downstream/modules/platform/ref-controller-database-settings.adoc
```diff
@@ -15,14 +15,16 @@ To improve the performance of the PostgreSQL server, configure the following _Gr
...
+[NOTE]
 ====
-If you are compiling Postgres against OpenSSL 3.2, your system regresses to remove the parameter for User during startup. You can rectify this by using the BIO_get_app_data call instead of open_get_data. Only an administrator can make these changes, but it impacts all users connected to the PostgreSQL database. f you update your systems without the OpenSSL patch, you are not impacted, and you do not need to take action.
+You must restart the database server after changing the value for `shared_buffers`.
 ====
 
 [NOTE]
 ====
-You must restart the database server after changing the value for `shared_buffers`.
+If you are compiling Postgres against OpenSSL 3.2, your system regresses to remove the parameter for User during startup. You can rectify this by using the BIO_get_app_data call instead of open_get_data. Only an administrator can make these changes, but it impacts all users connected to the PostgreSQL database.
```
jself-sudoku, line 25: Hi Anni. Thanks for taking a look. Looks like you still need to apply the back ticks in this instance: 
```suggestion
If you are compiling Postgres against OpenSSL 3.2, your system regresses to remove the parameter for User during startup. You can rectify this by using the `BIO_get_app_data` call instead of `open_get_data`. Only an administrator can make these changes, but it impacts all users connected to the PostgreSQL database.
```

jself-sudoku:
LGTM

PR #1802, opened 2024-08-30
Add procedure to set registry variables

ianf77:
Add to docs to create a token at access.redhat.com/terms-based-registry/accounts is necessary for entering a registry_username and registry_password

https://issues.redhat.com/browse/AAP-29120

PR #1803, opened 2024-08-30
Add procedure to set registry variables (#1802)

ianf77:
https://issues.redhat.com/browse/AAP-29120
* Add procedure to set registry variables


ariordan-redhat (approved):
LGTM

PR #1816, opened 2024-09-03
AAP-29296 removed subscription info

emurtoug:
AAP-29296 

Commented out subscription info from the planning guide and archived assembly-attaching-subscriptions.adoc


dcdacosta (approved):
LGTM!

PR #1827, opened 2024-09-05
improving user orientation and some minor style tweaks (#1826)

g-murray:
[AAP-30535](https://issues.redhat.com/browse/AAP-30535)

Most procedures are missing user orientation, such as "Log into OpenShift Container Platform" and "select your AAP deployment" Adding these additional steps in to improve UX.


ianf77 (approved):
LGTM

PR #1832, opened 2024-09-05
Add info on configuring load balancer for container installer (#1831)

michellemacrh:
Containerized installer: There's no guide to configure loadbalancer for containerized installer

https://issues.redhat.com/browse/AAP-28700


hherbly (approved):
LGTM

PR #1878, opened 2024-09-12
Remove field from procedure

ianf77:
Remove entry from System settings

https://issues.redhat.com/browse/AAP-30850


hherbly (approved):
LGTM

PR #1916, opened 2024-09-16
2.3 Remove attributes from title names

ariordan-redhat:
2.3 Remove attributes from title names because they're causing build failures in Pantheon.


michellemacrh:
LGTM

ccopelloRH:
LGTM

PR #1895, opened 2024-09-15
AAP-31307 Add attributes for title URLs

ariordan-redhat:
Create attributes for docs links so that external links to other docs in the AAP docset don't change when titles are updated.
Backport to 2.5

PR #1920, opened 2024-09-16
2.5 AAP-31307 Add attributes for title URLs (#1895)

ariordan-redhat:
Create attributes for docs links so that external links to other docs in the AAP docset don't change when titles are updated.


rogrange (approved):
All punctuation looks consistent. Testing inline will be necessary. 

Review completed 16 Sep. 2024.

PR #1926, opened 2024-09-16
AAP-30395 Fix JSON example and update devtools attribute

ariordan-redhat:
https://issues.redhat.com/browse/AAP-30395
AAP-30395 Fix JSON example and update devtools attribute


dcdacosta (approved):
LGTM!

PR #1941, opened 2024-09-17
Adding custom inventory script example

EMcWhinn:
Improved custom inventory script documentation for Controller

https://issues.redhat.com/browse/AAP-11874

### DIFF
--- a/downstream/modules/platform/ref-controller-scm-inv-source-fields.adoc
+++ b/downstream/modules/platform/ref-controller-scm-inv-source-fields.adoc
@@ -36,6 +36,12 @@ You can perform an inventory update while a related job is running.
 
 {ControllerNameStart} uses the `ansible-inventory` module from Ansible to process inventory files, and supports all valid inventory syntax that {ControllerName} requires.
 
+[IMPORTANT]
+====
+You do not need to write inventory scripts in Python.
+You can enter any executable file in the source field and must run `chmod +x` for that file and check it into Git.
+====
+
 The following is a working example of JSON output that {ControllerName} can read for the import:
 
 ----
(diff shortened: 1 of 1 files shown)


downstream/modules/platform/ref-controller-scm-inv-source-fields.adoc
```diff
@@ -35,3 +35,52 @@ You can perform an inventory update while a related job is running.
 == Supported File Syntax
 
 {ControllerNameStart} uses the `ansible-inventory` module from Ansible to process inventory files, and supports all valid inventory syntax that {ControllerName} requires.
+
+The following is an example of a working example of JSON output that {ControllerName} can read in:
```
tvo318, line 39:
```suggestion
The following is a working example of JSON output that {ControllerName} can read for the import:
```

downstream/modules/platform/ref-controller-scm-inv-source-fields.adoc
```diff
@@ -35,3 +35,52 @@ You can perform an inventory update while a related job is running.
...
+            "host4",
+            "host5",
+            "host6",
+            "host7",
+            "host8",
+            "host9"
+        ]
+    }
+} 
+----
```
tvo318, line 86:
```suggestion
----
For information on how to implement the inventory script, see support article: https://access.redhat.com/solutions/6997130.
```

tvo318 (changes requested):
Small comments for clarification as requested in the JIRA issue. Thanks for doing this!

AlanCoding:
Here's examples of inventory files:

https://github.com/ansible/test-playbooks/tree/main/inventories

Here's an example of an inventory script inside of that:

https://github.com/ansible/test-playbooks/blob/main/inventories/changes.py

But a big non-obvious part of this is that it's an executable file, meaning you need to run `chmod +x` for that file and then check it into git. You can see Github indicate this with:

Screenshot from 2024-09-17 13-31-56

The wording here isn't correct, but I want to be clear about what the JSON format means. That's what the inventory script would write to standard out. There is a standard format for this, and it's the same as what `ansible-inventory` outputs.

Inventory scripts don't even necessarily need to be written in python. They can be any executable, and that level of generality is why this is often hard to follow.

EMcWhinn:
Thanks @tvo318 and @AlanCoding for the extra context here. I've added an "IMPORTANT" admonition and some additional resources. Does this now look okay or is there anything else important missing?

AlanCoding (approved):

michellemacrh:
LGTM!

PR #1942, opened 2024-09-17
AAP-30716 Add YAML separator in devtools example

ariordan-redhat:
AAP-30716 Add YAML separator in devtools example
Affects `titles/develop-automation-content/`


g-murray (approved):
LGTM

PR #1983, opened 2024-09-18
opg-fix - Removed activating manifest files 

dcdacosta:
This PR removes the activating manifest files from the Operations guide because the content as moved to the Access management guide.


hherbly (approved):
LGTM

PR #1984, opened 2024-09-18
AAP-31577 Add attribute for unified platform dashboard

ariordan-redhat:
AAP-31577 Add attribute for unified platform dashboard
Backport to 2.5 only

### DIFF
--- a/downstream/attributes/attributes.adoc
+++ b/downstream/attributes/attributes.adoc
@@ -17,6 +17,7 @@
 :VMBase: VM-based installation
 :OperatorBase: operator-based installation
 :ContainerBase: container-based installation
+:PlatformDashboard: platform dashboard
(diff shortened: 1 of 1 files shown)


downstream/attributes/attributes.adoc
```diff
@@ -17,6 +17,7 @@
 :VMBase: VM-based installation
 :OperatorBase: operator-based installation
 :ContainerBase: container-based installation
+:PlatformUI: platform dashboard
```
dcdacosta, line 20: Hey @ariordan-redhat ! I think you might want to call this PlatformDashboard because there is discussion around terminology for using platform UI instead of platform gateway in the future and as named, this might cause confusion. Otherwise, I think this looks good.

PR #1988, opened 2024-09-18
2.5 AAP-31577 Add attribute for unified platform dashboard (#1984)

ariordan-redhat:
AAP-31577 Add attribute for unified platform dashboard


dcdacosta (approved):
LGTM!

PR #2003, opened 2024-09-20
Update inventory file vars appendix

michellemacrh:
Addition of the following vars sections:
- ref-images-inventory-variables.adoc
- ref-database-inventory-variables.adoc
- ref-receptor-inventory-variables.adoc

Used this [Inventory var diff spreadsheet](https://docs.google.com/spreadsheets/d/1wSZiIoDu-_uRY8l1R1vYQlOG5PPrK4MrP3jla8701KA/edit?gid=413143671#gid=413143671)

Update product docs installer var appendix with 2.5 variables

https://issues.redhat.com/browse/AAP-30775


downstream/modules/platform/ref-database-inventory-variables.adoc
```diff
@@ -0,0 +1,67 @@
...
+
+[cols="50%,50%,50%",options="header"]
+|====
+| *RPM variable name* | *Container variable name* | *Description*
+| `pg_ssl_mode` | | Choose one of the two available modes: `prefer` and `verify-full`. 
+
+Set to `verify-full` for client-side enforced SSL. 
+
+Default = `prefer`
+| `postgres_ssl_cert` | `postgresql_tls_cert` | Location of the PostgreSQL SSL/TLS certificate.
```
g-murray, line 13: I think SSL/TSL are two different certificates? If that is the case try to avoid using forward slashes when referring to both, in most cases "or" or "and" slot in easy enough. Below is from the IBM style guide:

"In most cases, do not use slashes in text because they can be ambiguous. For example, 1/2 can mean one half, one and two, one of two, and one or two. Rewrite the text using and, of, or or."

g-murray (approved):
One small  style comment, but besides that LGTM

downstream/modules/platform/ref-database-inventory-variables.adoc
michellemacrh: Oh yeah, I never know with this one because it conflicts with this statement: 
Screenshot 2024-09-20 at 10 14 04

g-murray: Ah if that is in our documentation, I feel that's a fair exception so!

PR #2040, opened 2024-09-23
AAP-22994-0922 Added missing part of 3rd bullet for EDA controller sys requirments

jself-sudoku:
Adding an important part of the EDA controller system requirements to `downstream/modules/platform/ref-eda-system-requirements.adoc`


g-murray (approved):
LGTM

PR #2043, opened 2024-09-23
Add install-related note to operator install doc

g-murray:
[AAP-31823](https://issues.redhat.com/browse/AAP-31823)

Adding the following 'Important note' to both install chapters (chapter 2 and 3):

"You can only install a single instance of the {OperatorPlatform} into a single namespace. 
Installing multiple instances in the same namespace can lead to improper operation for both operator instances."

Also made some minor tweaks to word choice and active voice.


ianf77 (approved):
Just a small comment, In the past we've had to remove the word 'modal' because no-one knew what it was meant to be.

PR #2048, opened 2024-09-23
EDA Controller & EDA Controller Variables Update

annibond:
- updated platform/ref-eda-controller-variables.adoc & platform/ref-controller-variables.adoc
 
 - docs.google.com/spreadsheets/d/1wSZiloDu-_uRY8l1E1vYQlO5PPrK4MrP3jla8701K/edit?gid=143143671

PR #2051, opened 2024-09-23
EDA Controller & EDA Controller Variables Update (#2048)

annibond:
* updated eda controller & gateway variables

* Update ref-eda-controller-variables.adoc

Changed EDA to {EDAName}


dcdacosta (approved):
LGTM!

PR #2056, opened 2024-09-23
tbl-fix - Fixing the table formatting in EDA Variables

dcdacosta:
This PR fixes the table formatting issue in the EDA Variables information.


hherbly (approved):
LGTM!

PR #2028, opened 2024-09-22
AAP-30402 Set docker path in dev containers extension

ariordan-redhat:
AAP-30402 Set docker path to `docker` or `podman` in dev containers extension

Affects `/titles/develop-automation-content/`

PR #2067, opened 2024-09-24
2.5 AAP-30402 Set docker path in dev containers extension (#2028)

ariordan-redhat:
Backports #2028 

AAP-30402 Set docker path to docker or podman in dev containers extension

Affects /titles/develop-automation-content/


g-murray (approved):
LGTM

PR #2078, opened 2024-09-24
AAP-30714 Devtools container package manager is microdnf

ariordan-redhat:
AAP-30714 The devtools container is based on a UBI minimal image, which uses microdnf as a package manager, not yum or dnf.

Affects `titles/develop-automation-content`


annibond (approved):
lgtm! very easy to read/get

PR #2084, opened 2024-09-24
AAP-30714 Devtools container package manager is microdnf (#2078)

ariordan-redhat:
Backports #2078 to 2.4

AAP-30714 The devtools container is based on a UBI minimal image, which uses microdnf as a package manager, not yum or dnf.

Affects titles/develop-automation-content


michellemacrh (approved):
LGTM

PR #2086, opened 2024-09-24
AAP-31856 Added the HS assembly to Containerized Install Guide

jself-sudoku:
Add new topic on Horizontal Scaling that was already added to the RPM Installation Guide to the Containerized Installation Guide. 

Note to @michellemacrh : Since this topic is in it's own assembly and will likely be expanded to include procedures to use HS in separate subsections for controller, hub, and possibly mesh, I added the HS assembly right after the first assembly.


michellemacrh:
LGTM

PR #2098, opened 2024-09-25
2.5 Update abstract of Containerized installation guide

michellemacrh:
Backporting the change of abstract to 2.5 so it's correct for GA.


g-murray (approved):
LGTM

PR #2111, opened 2024-09-25
AAP-31874 - Reconciling chapter 3 of the Access management guide

dcdacosta:
This PR reconciles chapter 3 of the Access management guide against the post code-freeze test environment and includes minor edits and late SME review changes.

### DIFF
--- a/downstream/modules/platform/con-gw-authenticator-map-triggers.adoc
+++ b/downstream/modules/platform/con-gw-authenticator-map-triggers.adoc
@@ -19,7 +19,7 @@ If you are only keying off one group it doesn’t matter if you select *"and"* o
 +
 * *Groups:* This is a list of one or more groups coming from the authentication system that the user must be a member of. See the Operation field to determine the behavior of the trigger if more than one group is specified in the trigger.
 
-Attribute:: The map is true or false based on a users attributes coming from the source system.When defining an attribute trigger, the authentication mapping expands to include the following selections:
+Attribute:: The map is true or false based on a users attributes coming from the source system. When defining an attribute trigger, the authentication mapping expands to include the following selections:
 +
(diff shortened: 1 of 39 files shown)


downstream/modules/platform/con-gw-authenticator-map-triggers.adoc
annibond, line 1: Line 22 misses a space between a period and the next sentence: 

Attribute:: The map is true or false based on a users attributes coming from the source system.When defining an attribute trigger, the authentication mapping expands to include the following selections:

downstream/modules/platform/proc-gw-configure-auth-details.adoc
annibond, line 1: On line 38, those should be two sentences rather than one. The second one would begin at "The authenticator maps will not have an opinion.."

downstream/modules/platform/con-gw-authenticator-map-triggers.adoc
annibond, line 1: Line 22 misses a space between a period and the next sentence:

Attribute:: The map is true or false based on a users attributes coming from the source system.When defining an attribute trigger, the authentication mapping expands to include the following selections:

annibond (changes requested):
two small changes

downstream/modules/platform/proc-gw-configure-auth-details.adoc
dcdacosta: Fixed.

downstream/modules/platform/con-gw-authenticator-map-triggers.adoc
dcdacosta: Fixed.

annibond (approved):

PR #2143, opened 2024-09-26
Update URL attribute to use html instead of html-single (#2118)

ariordan-redhat:
URL attributes were using `html-single` for docs.redhat.com (ie whole doc is shown) instead of `html` (multi-page).
Previous links used multi-page.
However, the URLs for assemblies / modules change depending on whether html or html-single is used in a link.

For example

**multi-page (`html`):**
https://docs.redhat.com/en/documentation/red_hat_ansible_automation_platform/2.4/html/automation_controller_user_guide/assembly-controller-applications#proc-controller-apps-create-tokens

**single-page (`html-single`):**
https://docs.redhat.com/en/documentation/red_hat_ansible_automation_platform/2.4/html-single/automation_controller_user_guide/index#proc-controller-apps-create-tokens

All of our links were configured for all releases up to and including 2.4 for the multi-page version.
This PR will make it easier to convert links to use the doc URL attribute - the link within the doc won't need to be updated.

### DIFF
--- a/downstream/attributes/attributes.adoc
+++ b/downstream/attributes/attributes.adoc
@@ -285,155 +285,155 @@
 //
 // titles/troubleshooting-aap
 :TitleTroubleshootingAAP: Troubleshooting Ansible Automation Platform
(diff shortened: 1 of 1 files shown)


dcdacosta (approved):
LGTM!

PR #2156, opened 2024-09-27
Fix broken xrefs in Managing automation content

ariordan-redhat:
Fixes `xref`s that were causing the Managing automation content build to fail in Pantheon.

Affects `hub/managing-content/`


emurtoug (approved):
LGTM

PR #2159, opened 2024-09-27
2.5 Fix broken xrefs in Managing automation content (#2156)

ariordan-redhat:
Fixes `xref`s that were causing the Managing automation content build to fail in Pantheon.

Affects `hub/managing-content/`


emurtoug (approved):
LGTM

PR #2160, opened 2024-09-27
2.5 links fix

g-murray:
Just fixing one link that reference the old title name, and tidying up two others with a more accurate description


michellemacrh:
LGTM!

PR #2201, opened 2024-09-30
2.5 Remove reference to controller UI in troubleshooting guide (#2199)

michellemacrh:
Comment out a troubleshooting section that is relevant to the controller UI which should no longer be accessed.


ianf77 (approved):
LGTM

PR #2213, opened 2024-09-30
Fix xref in Controller Admin

ianf77:
Should be a link

Fix reference in Controller Admin Guide

https://issues.redhat.com/browse/AAP-32256


EMcWhinn (approved):
LGTM

PR #2229, opened 2024-09-30
Fixing assembly ID to be unique in Organizations content

dcdacosta:
Changed generic template assembly id [assembly-my-user-story_{context}] to something unique.


hherbly (approved):
LGTM!

PR #2241, opened 2024-10-01
AAP-29062 RHDH Add requirement for developer hub portal account

ariordan-redhat:
https://issues.redhat.com/browse/AAP-28962

Add optional requirement for a developer gub account so that users can access the learning paths linked from the Ansible plug-ins on RHDH.

Affects `titles/aap-plugin-rhdh-using/`

### DIFF
--- a/downstream/assemblies/devtools/assembly-rhdh-using.adoc
+++ b/downstream/assemblies/devtools/assembly-rhdh-using.adoc
@@ -11,6 +11,13 @@ From the {RHDH} UI, you can navigate to your {PlatformNameShort} instance, where
 This document describes how to use the {AAPRHDH}.
 It presents a worked example of developing a playbook project for automating updates to your firewall configuration on RHEL systems.
 
+== Optional requirement
+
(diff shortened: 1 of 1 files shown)


downstream/assemblies/devtools/assembly-rhdh-using.adoc
```diff
@@ -11,6 +11,13 @@ From the {RHDH} UI, you can navigate to your {PlatformNameShort} instance, where
 This document describes how to use the {AAPRHDH}.
 It presents a worked example of developing a playbook project for automating updates to your firewall configuration on RHEL systems.
 
+== Optional requirement
+
+The {AAPRHDH} link to Learning Paths on the Red Hat developer portal,
```
g-murray, line 16: I don't know how strict the "Red{nbsp}Hat" rule, but the style guide does state that any instance of Red Hat should have a non breaking space so it doesn't run over two lines and separate due to formatting. I'll let you decide if you want to follow it or not as it's kind of a silly rule.

PR #2353, opened 2024-10-15
Updating OCP version

g-murray:
[AAP-33381](https://issues.redhat.com/browse/AAP-33381)

Updating the OCP version that is compatible with AAP OP 2.5


jself-sudoku:
LGTM

PR #2355, opened 2024-10-15
Add note about external server (#2352)

ianf77:
Add explict statement stating 'NFS is external server'

https://issues.redhat.com/browse/AAP-32889


ariordan-redhat (approved):
LGTM

PR #2372, opened 2024-10-17
fixing saas links and attribute

g-murray:
Quick fix to incorrect links and a few other minor tweaks, as called out in the Oct 17th Hackathon,


ariordan-redhat (approved):
LGTM

PR #2373, opened 2024-10-17
Clouds - update attributes and boilerplate text

ariordan-redhat:
Update the boilerplate text in Azure and SaaS docs, and add attributes for AAP menus and AAP docs URLs.


g-murray (approved):
LGTM

PR #2375, opened 2024-10-17
Clouds - Archive AWS and GCP docs

ariordan-redhat:
Archive titles, stories, and topics relating to AAP on AWS and AAP on GCP


g-murray (approved):
LGTM

PR #2378, opened 2024-10-18
AAP-32794 updates

g-murray:
[AAP-32794](https://issues.redhat.com/browse/AAP-32794) 

Chapter 3. Installing Ansible Automation Platform Operator from the OpenShift Container Platform CLI in the new Ansible Automation Platform 2.5 documentation does not configure the AnsibleAutomationPlatform object as required for the new deployment.

### DIFF
--- a/downstream/modules/platform/proc-install-cli-aap-operator.adoc
+++ b/downstream/modules/platform/proc-install-cli-aap-operator.adoc
@@ -55,40 +55,63 @@ spec:
   source: redhat-operators
   sourceNamespace: openshift-marketplace
 ---
-apiVersion: automationcontroller.ansible.com/v1beta1
-kind: AutomationController
-metadata:
-  name: example
-  namespace: ansible-automation-platform
-spec:
-  replicas: 1
-
 -----
 +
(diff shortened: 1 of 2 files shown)


downstream/modules/platform/proc-install-cli-aap-operator.adoc
```diff
@@ -55,40 +55,63 @@ spec:
...
-  name: __<automation_controller_name>__
-  namespace: ansible-automation-platform
------
 . Run the [command]`*oc apply*` command to create the objects specified in the [filename]`sub.yaml` file:
 +
 -----
 oc apply -f sub.yaml
 -----
++
+. Verify the CSV PHASE reports Succeeded before proceeding using the [command]`oc get csv -n ansible-automation-platform` command:
```
michellemacrh, line 68: Should "Succeeded" be "succeeded" (capitalization)?

g-murray: I was mirroring it as it appears in the code block, so actually I should have used `Succeeded` I think actually, do you agree with that?

michellemacrh: Ah fair, I wonder if it's referencing the code block directly should it be denoted with quotes or something!?

michellemacrh:
LGTM

PR #2398, opened 2024-10-21
2.4 Fix  title attribute

ariordan-redhat:
Fix the attributes that reference the Controller getting started guide.


dcdacosta (approved):
LGTM!

PR #2402, opened 2024-10-22
Add conditional text to Horizontal scaling in EDA

michellemacrh:
Add conditional content to con-hs-eda-controller.adoc to account for the differences between RPM and Container installations.

Correct inventory file attributes for containerized installer in horizontal scaling content

https://issues.redhat.com/browse/AAP-33624

### DIFF
--- a/downstream/modules/platform/con-hs-eda-controller.adoc
+++ b/downstream/modules/platform/con-hs-eda-controller.adoc
@@ -7,9 +7,11 @@ With {EDAcontroller}, you can set up horizontal scaling for your events automati
 The following node types are used in this deployment:
 
 API node type:: Responds to the HTTP REST API of {EDAcontroller}.
-Worker node type:: Runs an {EDAName} worker, which is the component of Event-Driven Ansible that not only manages projects and activations, but also executes the activations themselves.
+Worker node type:: Runs an {EDAName} worker, which is the component of {EDAName} that not only manages projects and activations, but also executes the activations themselves.
 Hybrid node type:: Is a combination of the API node and the worker node.
 
+// This content is used in RPM installation
+ifdef::aap-install[]
 The following example shows how you would set up an inventory file for horizontal scaling of {EDAcontroller} on {RHEL} VMs using the host group name `[automationedacontroller]` and the node type variable `eda_node_type`:
 
 -----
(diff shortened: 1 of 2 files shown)


downstream/modules/platform/con-hs-eda-controller.adoc
```diff
@@ -7,9 +7,11 @@ With {EDAcontroller}, you can set up horizontal scaling for your events automati
 The following node types are used in this deployment:
 
 API node type:: Responds to the HTTP REST API of {EDAcontroller}.
-Worker node type:: Runs an {EDAName} worker, which is the component of Event-Driven Ansible that not only manages projects and activations, but also executes the activations themselves.
+Worker node type:: Runs an {EDAName} worker, which is the component of {EDAName} that not only manages projects and activations, but also executes the activations themselves.
 Hybrid node type:: Is a combination of the API node and the worker node.
 
+// This content is used in RPM installation
+ifdef::aap-install[]
 The following example shows how you would set up an inventory file for horizontal scaling of {EDAcontroller} on {RHEL} VMs using the host group name `[automationedacontroller]` and the node type variable `eda_node_type`:
```
g-murray, line 15: Not a deal breaker by any means, but I think we are encouraged to avoid conditionals. You could sub in something like "can"?

downstream/modules/platform/con-hs-eda-controller.adoc
```diff
@@ -21,3 +23,19 @@ routable_hostname=automationedacontroller-api.example.com eda_node_type=api
 # worker node
 3.88.116.112 routable_hostname=automationedacontroller-api.example.com eda_node_type=worker
 -----
+endif::aap-install[]
+
+// This content is used in Containerized installation
+ifdef::container-install[]
+The following example shows how you would set up an inventory file for horizontal scaling of {EDAcontroller} on {RHEL} VMs using the host group name `[automationeda]` and the node type variable `eda_node_type`:
```
g-murray, line 30: same comment on conditional tense

g-murray (approved):
LGTM

downstream/modules/platform/con-hs-eda-controller.adoc
michellemacrh: I didn't write the original content (I was more focused on the ifdef conditionals) but I'll update to reflect the feedback 👍

PR #2414, opened 2024-10-23
Added a short note denoting the Python 3.5 removal as a supported (#2…

rogrange:
…413)

remote version only applies to Ansible version 2.17.

Resolves: AAP-33894


hherbly (approved):
LGTM!

PR #2417, opened 2024-10-23
AAP-34122: moves assembly from upgrading 2.5 to 2.5x to Installing on OCP guide

hherbly:
See [AAP-34122](https://issues.redhat.com/browse/AAP-34122) for more info.


dcdacosta (approved):
LGTM!

PR #2418, opened 2024-10-23
AAP-30120 Added extra bullet to the admonition in Setting up a rulebook activation section

jself-sudoku:
Added an extra point to the admonition in the proc-eda-set-up-rulebook-activation.adoc module. Alerts users/customers to only select 1 RH AAP credential when setting up a rulebook activation using the RH AAP credential.


zkayyali812 (approved):

hherbly (approved):
LGTM!

PR #2421, opened 2024-10-23
Backport-2.5-2418 AAP-30120 Added extra bullet to the admonition in Setting up a rulebook activation

jself-sudoku:
Added an extra point to the admonition in the proc-eda-set-up-rulebook-activation.adoc module. Alerts users/customers to only select 1 RH AAP credential when setting up a rulebook activation using the RH AAP credential.

### DIFF
--- a/downstream/modules/eda/proc-eda-set-up-rulebook-activation.adoc
+++ b/downstream/modules/eda/proc-eda-set-up-rulebook-activation.adoc
@@ -24,9 +24,10 @@ Credential:: Select 0 or more credentials for this rulebook activation. This fie
 +
 [NOTE]
 ====
-The credentials that display in this field are customized based on your rulebook activation and only include the following credential types: Vault, {PlatformName}, or any custom credential types that you have created. For more information about credentials, see xref:eda-credentials[Credentials].
+* If you plan to use a {PlatformName} credential, you can _only_ select 1 of this credential type for your rulebook activation.
+* The credentials that display in this field are customized based on your rulebook activation and only include the following credential types: Vault, {PlatformName}, or any custom credential types that you have created. For more information about credentials, see xref:eda-credentials[Credentials].
 ====
-//[J. Self] Might need to update the link above for the updated Credentials section.
+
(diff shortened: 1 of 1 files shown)


downstream/modules/eda/proc-eda-set-up-rulebook-activation.adoc
```diff
@@ -24,9 +24,10 @@ Credential:: Select 0 or more credentials for this rulebook activation. This fie
 +
 [NOTE]
 ====
-The credentials that display in this field are customized based on your rulebook activation and only include the following credential types: Vault, {PlatformName}, or any custom credential types that you have created. For more information about credentials, see xref:eda-credentials[Credentials].
+* If you plan to use a {PlatformName} credential, you can _only_ select 1 of this credential type for your rulebook activation.
```
dcdacosta, line 27: One of "this" credential type? Is it a specific type, if so it might be better to specify which type it is. Otherwise, I would reword this to say:

 If you plan to use a {PlatformName} credential, you can _only_ select 1 credential type for your rulebook activation.

dcdacosta (commented):
Minor suggestion to eliminate possible confusion.

downstream/modules/eda/proc-eda-set-up-rulebook-activation.adoc
jself-sudoku: I rewrote this sentence in different ways to reduce redundancy or make things as clear as possible. By starting the sentence with "If you plan to use a Red Hat AAP credential...", I thought it would be clear that they should only use 1 of that specific credential. I think the sentence you suggested, while seemingly more concise, might still be confusing because they can actually select multiple credentials for the Credential field. I think I should move this bullet point to the second bullet and let the original point be the first point because it explains the available options. At any rate, I would say proceed with the merge here and the tweak will be handled in a different PR.

dcdacosta: Got it, I'll proceed with the merge. :D

dcdacosta (approved):
LGTM!

PR #2425, opened 2024-10-24
2.5 Update RPM-B-ENV-B topology for Redis (#2424)

michellemacrh:
The mixed enterprise topology needs to be updated to account for 3 VMs for gateway and eda

Topology updates for RPM Redis support

https://issues.redhat.com/browse/AAP-32331


hherbly (approved):
lgtm!

PR #2487, opened 2024-10-30
Added line to clarify Python version removal.

rogrange:
Added the Python versions that have been removed as supported remote versions.

Resolves: #2414
See also: AAP-33894


sayjadha (commented):
LGTM.

PR #2489, opened 2024-10-30
This fixes the naming convention for the attribute URLUpgrade

rogrange:
Fixes the naming convention for the attribute from upgrade_migration
to
rpm_upgrade_migration

Resolves: AAP-34824


hherbly (approved):
LGTM!

PR #2502, opened 2024-10-31
AAP-29905B Correction to note to link user to replacing tokens with RHAAP creds

jself-sudoku:
Updated the note in downstream/assemblies/eda/assembly-eda-set-up-rhaap-credential.adoc (follows Google draft).


hherbly (approved):
LGTM!

PR #2509, opened 2024-11-04
Applying updates from AAP-35076

g-murray:
The steps are very vague. Adding more detailed steps and an example to follow.

### DIFF
--- a/downstream/modules/platform/proc-operator-create_crs.adoc
+++ b/downstream/modules/platform/proc-operator-create_crs.adoc
@@ -2,15 +2,44 @@
 
 = Creating {PlatformNameShort} custom resources
 
-Create an {PlatformNameShort} custom resource (CR) that specifies the names of your existing deployments, in the same namespace.
+After upgrading to the latest version of {OperatorPlatformNameShort} on {OCPShort}, you can create an {PlatformNameShort} custom resource (CR) that specifies the names of your existing deployments, in the same namespace.
 
 .Procedure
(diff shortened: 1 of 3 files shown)


downstream/modules/platform/proc-operator-create_crs.adoc
```diff
@@ -19,3 +48,9 @@ This enables upgrading a specific deployment, like a controller, without updatin
 
 The recommended approach however, is to upgrade the operator and use the default image values. 
 ====
+
+.Verification
+Go to your {OperatorPlatformNameShort} deployment and click btn:[All instances] to verify if all instances deployed correctly.
```
emurtoug, line 53: You might consider using "Navigate to" rather than "Go to" here. Up to you, though!

downstream/modules/platform/proc-operator-create_crs.adoc
```diff
@@ -19,3 +48,9 @@ This enables upgrading a specific deployment, like a controller, without updatin
...
+You should see the *{PlatformNameShort}* instance and the deployed *AutomationController*, *EDA*, and *AutomationHub* instances here.
+
+Alternatively you can check by the command line, run: `oc get route`
```
emurtoug, line 56: "Alternatively, you can verify whether all instances deployed correctly by running `oc get route` in the command line." This might be a clearer statement

downstream/modules/platform/proc-operator-create_crs.adoc
emurtoug: "to verify whether all instances have deployed correctly." This might be slightly clearer.

PR #2522, opened 2024-11-05
Clouds - Add attributes for current AAP version on Azure

ariordan-redhat:
Self-managed Azure is not upgrading to the new AAP release for a while, so we need an attribute to track the current Azure version.


g-murray (approved):
LGTM

PR #2224, opened 2024-09-30
AAP-32141a Updates to RHDH Install doc

ariordan-redhat:
AAP-32141a Updates to RHDH Install doc

Affects `titles/aap-plugin-rhdh-install`

PR #2149, opened 2024-09-26
AAP-32141 Add Installing on Operator chapter to RHDH install guide

ariordan-redhat:
AAP-32141 Add Installing on Operator chapter to RHDH install guide

Affects titles/aap-plugin-rhdh-install/

Preview here: https://ariordan-redhat.github.io/aap-builds/aap-plugin-rhdh-install-scaffold-plugin-operator-install-2024-09-26.html

PR #2539, opened 2024-11-07
AAP-32141 Add Installing on Operator chapter to RHDH install guide (#2149) (#2224)

ariordan-redhat:
Backports #2149 and #2224 to 2.5
Affects `titles/aap-plugin-rhdh-install`


g-murray (approved):
LGTM

PR #2541, opened 2024-11-07
AAP-35359 Clouds - update sync script for AWS / GCP deprecation

ariordan-redhat:
AAP-35359 Clouds - update sync script for AWS / GCP deprecation

The AWS and GCP docs have been deprecated. Update the sync script so that Jenkins does not attempt to sync these to the production repo.


g-murray (approved):
LGTM

PR #2523, opened 2024-11-06
Updating "pods/log" resource

EMcWhinn:
Revamp the permissions for "pods/log" resource of an AAP container group

https://issues.redhat.com/browse/AAP-14596

Affects `titles/controller-user-guide`

PR #2551, opened 2024-11-11
Updating "pods/log" resource (#2523)

EMcWhinn:
Revamp the permissions for "pods/log" resource of an AAP container group

https://issues.redhat.com/browse/AAP-14596


michellemacrh:
LGTM

PR #2514, opened 2024-11-05
Added not about receptor not listening

ianf77:
Explain in documentation that receptor is not listening on first controller

https://issues.redhat.com/browse/AAP-23036

PR #2558, opened 2024-11-12
Added note about receptor not listening (#2514)

ianf77:
* Added not about receptor not listening

Explain in documentation that receptor is not listening on first controller

https://issues.redhat.com/browse/AAP-23036

* Added note about receptor not listening

Explain in documentation that receptor is not listening on first controller

https://issues.redhat.com/browse/AAP-23036


EMcWhinn (approved):
LGTM

PR #2561, opened 2024-11-12
AAP-34041 plug-ins: view container logs on OCP UI

ariordan-redhat:
Describe how to view container logs for a pod from the OpenShift UI so that the user can verify that the plug-ins have been installed.
Affects `titles/aap-plugin-rhdh-install/`


hherbly (approved):
LGTM!

PR #2563, opened 2024-11-12
Updating awx commands to gateway

EMcWhinn:
OAuth2 product docs reference awx-manage instead of aap-gateway-manage

https://issues.redhat.com/browse/AAP-35735

Affects `titles/central-auth`


michellemacrh:
LGTM!

PR #2569, opened 2024-11-13
AAP-35811 Azure doc cleanup before 2.5 work starts

ariordan-redhat:
AAP-35811 Azure doc cleanup before 2.5 work starts


g-murray (approved):
LGTM

PR #2579, opened 2024-11-15
AAP-33304: Document case-sensitivity of authenticator map group triggers

BrennanPaciorek:


downstream/modules/platform/con-gw-authenticator-map-triggers.adoc
```diff
@@ -18,7 +18,12 @@ If you are only keying off one group it doesn’t matter if you select *"and"* o
 ====
 +
 * *Groups:* This is a list of one or more groups coming from the authentication system that the user must be a member of. See the *Operation* field to determine the behavior of the trigger if more than one group is specified in the trigger.
-
++
+[NOTE]
+====
+You will need to provide group identifiers in the same casing that authenticator backends return them to clients, as this field is case-sensitive. (Example, cn=johnsmith,dc=example,dc=com instead of CN=johnsmith,DC=example,DC=com)
```
dcdacosta, line 24: Suggest rewording:

Group identifiers are case-sensitive and must match the authenticator backend. For example, `cn=johnsmith, dc=example, dc=com` instead of CN=johnsmith, DC=example, DC=com`.

BrennanPaciorek: Done.

dcdacosta (approved):
LGTM!

PR #2583, opened 2024-11-18
Editing sample service account details

EMcWhinn:
Update sample service account details

https://issues.redhat.com/browse/AAP-36116

Affects `titles/controller-user-guide`

PR #2585, opened 2024-11-18
Editing sample service account details (#2583)

EMcWhinn:
* Editing sample service account details

Update sample service account details

https://issues.redhat.com/browse/AAP-36116

* Update per peer review


michellemacrh:
LGTM!

PR #2611, opened 2024-11-20
Add note to LDAP configuration regarding signed certificates

dcdacosta:
This PR adds a note regarding signed certificates to the LDAP configuration doc as requested by AAP-36326.

### DIFF
--- a/downstream/modules/platform/proc-controller-set-up-LDAP.adoc
+++ b/downstream/modules/platform/proc-controller-set-up-LDAP.adoc
@@ -6,6 +6,11 @@
 
 As a platform administrator, you can configure LDAP as the source for account authentication information for {PlatformNameShort} users. 
 
+[NOTE]
+====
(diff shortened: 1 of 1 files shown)


downstream/modules/platform/proc-controller-set-up-LDAP.adoc
```diff
@@ -6,6 +6,11 @@
 
 As a platform administrator, you can configure LDAP as the source for account authentication information for {PlatformNameShort} users. 
 
+[NOTE]
+====
+If the LDAP server you want to connect to has a certificate that is self-signed or signed by a corporate internal certificate authority (CA), the CA certificate must be added to the system’s trusted CAs. Otherwise, connection to the LDAP server will result in an error that the certificate issuer is not recognized.
```
hherbly, line 11: would it be reasonable/useful to link to some info on how to add certificates to the trusted CA list?

dcdacosta: Agree this would be helpful but I was unable to find the details. Created a ticket to research this further: https://issues.redhat.com/browse/AAP-36336

PR #2613, opened 2024-11-21
Added mTLS for postgres to containerized install

ianf77:
Added module, and added variables to appendix

Containerized installation - Add certificate auth support (mTLS) for PostgreSQL external DB

https://issues.redhat.com/browse/AAP-36273

### DIFF
--- a/downstream/modules/platform/ref-database-inventory-variables.adoc
+++ b/downstream/modules/platform/ref-database-inventory-variables.adoc
@@ -6,89 +6,89 @@
 |====
 | *RPM variable name* | *Container variable name* | *Description*
 //EDA database variables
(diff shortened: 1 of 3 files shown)


downstream/modules/platform/ref-database-inventory-variables.adoc
```diff
@@ -5,11 +5,122 @@
 [cols="50%,50%,50%",options="header"]
 |====
 | *RPM variable name* | *Container variable name* | *Description*
+//EDA database variables
+| `automationedacontroller_install_pg_host` | `eda_` | Configure the PostgreSQL host for {EDAName}, which can be an externally managed database.
```
michellemacrh, line 9: Just a suggestion but for the sake of clarity I'd propose listing out the full variable name as opposed to just `eda_` for example as this may be confusing for the user

PR #2621, opened 2024-11-22
Fixes ascii error in Note in Managing automation content guide

hherbly:
It looks like this error originated from a missed backport. The "main" version is as it should be.


dcdacosta (approved):
LGTM!

PR #2680, opened 2024-12-10
Removing 20 node limit from 2.4 doc

EMcWhinn:
[Docs] 20 node limit in docs is obselete

https://issues.redhat.com/browse/AAP-23909


michellemacrh:
LGTM!

PR #2711, opened 2024-12-17
AAP-33505 - Add clarifying note for SAML IDP extra_data details

dcdacosta:
This PR adds a note to the SAML setting for extra_data based on the information in https://issues.redhat.com/browse/AAP-33505.

### DIFF
--- a/downstream/modules/platform/proc-controller-set-up-SAML.adoc
+++ b/downstream/modules/platform/proc-controller-set-up-SAML.adoc
@@ -71,11 +71,17 @@ For more information and additional options, see link:https://github.com/SAML-To
 +
 [NOTE]
 ====
-Any parameters other than the defaults (User Last Name, User First Name, User Email, etc.) must include attributes for mapping in the *SAML IDP to extra_data attribute mapping* field. This ensures all non-default attributes are properly interpreted by Ansible Automation Platform. For example, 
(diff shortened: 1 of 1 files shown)


downstream/modules/platform/proc-controller-set-up-SAML.adoc
```diff
@@ -69,6 +69,16 @@ For more information and additional options, see link:https://github.com/SAML-To
...
+[NOTE]
+====
+Any parameters other than the defaults (User Last Name, User First Name, User Email, etc.) must include attributes for mapping in the *SAML IDP to extra_data attribute mapping* field. This ensures all non-default attributes are properly interpreted by Ansible Automation Platform. For example, 
+-----
+- Department
+- UserType
+- Organization
+-----
+====
++
```
john-westcott-iv, line 81:
```suggestion
[NOTE]
====
By default only the attributes in the SAML Assertion identified in the fields: User Email, Username, User Last Name, User First Name, and User Permanent ID are available for AAP authenticator maps.

If you would like to make other attributes from the SAML Assertion available for AAP authenticator maps you must include them in the *SAML IDP to extra_data attribute mapping* field. For example:
-----
- Department
- UserType
- Organization
-----

Alternatively, if you want all attributes returned by the SAML Assertion to be avialable for AAP authenticator_maps you can add the following setting to *Additional Authenticator Fields*:
-----
GET_ALL_EXTRA_DATA: true
-----

====
+
```

dcdacosta: SME changes added to source file with editorial modifications.

PR #2730, opened 2024-12-20
Added notice that OAuth and LDAP data migration is not supported in the platform ui

dcdacosta:
This PR adds the changes requested for the customer docs in https://issues.redhat.com/browse/AAP-37120. 

Once a KCS article is published to handle this via the API, the LDAP note will be update to include a link to the article.


hherbly (approved):
LGTM

PR #2737, opened 2025-01-03
Minor correction to container group procedure (#2736)

ianf77:
[Doc Bug] : Container group service account example missing.

https://issues.redhat.com/browse/AAP-38107


EMcWhinn (approved):
LGTM

PR #2738, opened 2025-01-03
Removing cleanup string from controller docs

EMcWhinn:
Remove docs on "Cleanup Expired OAuth 2 Tokens"

https://issues.redhat.com/browse/AAP-37805

Affects `controller-admin-guide`


ianf77 (approved):
LGTM

PR #2754, opened 2025-01-10
2.5 Update eda config procedures in Containerized installation (#2753)

michellemacrh:
The following files are specific to RPM installations and needed to be updated to account for container installations also:
- proc-add-eda-safe-plugin-var.adoc
- proc-hs-eda-setup.adoc

Add conditional statements to account for both RPM and CONT based procedures.

Affects titles:
- aap-containerized-install
- aap-installation-guide

Containerized installation - Ensure inventory file code snippets contain correct samples for RPM and CONT AAP

https://issues.redhat.com/browse/AAP-36654


g-murray (approved):
LGTM

PR #2759, opened 2025-01-13
AAP-38561 - Corrected superuser mapping content

dcdacosta:
This PR corrects the Superuser mapping content to correctly follow the superuser workflow rather than the Role workflow. It also corrects a UI selection in the Role mapping content.


jself-sudoku:
lgtm

PR #2762, opened 2025-01-14
Removed postgres message

ianf77:
Product Docs State That External Databases Are Not Supported

https://issues.redhat.com/browse/AAP-37869


ariordan-redhat (approved):
LGTM

PR #2766, opened 2025-01-14
Removed postgres message (#2762)

ianf77:
* Removed postgres message

Product Docs State That External Databases Are Not Supported

https://issues.redhat.com/browse/AAP-37869


michellemacrh:
LGTM!

PR #2781, opened 2025-01-15
AAP-36580 added a note that explicitly states 6 VMs are required

dcdacosta:
This PR updates the Redis doc to explicitly state the requirement for 6VMs and includes a link to the deployment topologies to address customer feedback in https://issues.redhat.com/browse/AAP-36580.


hherbly (approved):
LGTM

PR #2807, opened 2025-01-22
AAP-38856 updated note to specify that colocation limitation is specific to RPM

dcdacosta:
This PR updates the note to eliminate confusion regarding Redis colocation between RPM and containerized deployments.


hherbly (approved):
LGTM

PR #2809, opened 2025-01-23
Replaced missing variable

ianf77:
Replaced automationhub_main_url

Replace entry for automationhub_main_url

https://issues.redhat.com/browse/AAP-39218

### DIFF
--- a/downstream/modules/platform/ref-hub-variables.adoc
+++ b/downstream/modules/platform/ref-hub-variables.adoc
@@ -111,7 +111,7 @@ This parameter enables you to drive this configuration.
 
 | `automationhub_main_url` | `hub_main_url` | The main {HubName} URL that clients connect to.
 
-For example, https://<load balancer host>.
+For example, \https://<load balancer host>.
 
 Use `automationhub_main_url` to specify the main {HubName} URL that clients connect to if you are implementing {RHSSO} on your {HubName} environment.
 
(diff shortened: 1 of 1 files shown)


downstream/modules/platform/ref-hub-variables.adoc
```diff
@@ -109,6 +109,14 @@ Examples are `ansible-doc`, `ansible-lint`, and `flake8`.
 
 This parameter enables you to drive this configuration.
 
+| `automationhub_main_url` | `hub_main_url` | The main {HubName} URL that clients connect to.
```
michellemacrh, line 112: `hub_main_url` is also an entry further down in the file that should be removed to prevent duplication.

downstream/modules/platform/ref-hub-variables.adoc
```diff
@@ -109,6 +109,14 @@ Examples are `ansible-doc`, `ansible-lint`, and `flake8`.
...
+
+For example, https://<load balancer host>.
```
michellemacrh, line 114: It might be worth wrapping the url in code formatting so it doesn't resolve to a url for example `https://<load_balancer_host>`

michellemacrh (commented):
Some small suggestions :)

downstream/modules/platform/ref-hub-variables.adoc
ianf77: Escaped the entry instead, code formatting doesn't do anything.

michellemacrh:
LGTM

PR #2840, opened 2025-01-30
Update CONT-A to include RAM requirement for bundled install

michellemacrh:
- Add a note to Tested deployment models doc to state that 32 GB RAM is required for a bundled install for CONT-A
- Add the same note to Containerized installation doc
- Remove the previous warning that was in place as it's no longer relevant now we have the required RAM config
- Archive the warning snippet

Update Docs for hub_seed_collections=true for containerized installer

https://issues.redhat.com/browse/AAP-38830


downstream/assemblies/platform/assembly-aap-containerized-installation.adoc
```diff
@@ -56,6 +56,7 @@ Each virtual machine (VM) has the following system requirements:
 | Disk IOPS   | 3000    
 |====
 
+If performing a bundled installation of the growth topology with `hub_seed_collections=true`, then 32 GB RAM is recommended. Note that with this configuration, install times can take 45 minutes or more to complete.
```
elyezer, line 59: What about instead of:

> install times can take 45 minutes or more to complete

We write something like:

> install time is going to increate and may take 45 or more minutes alone to complete seeding the collections.

downstream/modules/topologies/ref-cont-a-env-a.adoc
```diff
@@ -11,6 +11,8 @@ image::cont-a-env-a.png[Container growth topology diagram]
 
 A single VM has been tested with the following component requirements: 16 GB RAM, 4 CPUs, 60 GB local disk, and 3000 IOPS. Resources, such as storage, can be increased based on the needs of the deployment.
 
+If performing a bundled installation with `hub_seed_collections=true`, then 32 GB RAM is recommended. Note that with this configuration, install times can take 45 minutes or more to complete.
```
elyezer, line 14: Is you like the previous suggestion, then we would need to also reflect it here

downstream/assemblies/platform/assembly-aap-containerized-installation.adoc
elyezer, line 59: What about instead of:

> install times can take 45 minutes or more to complete

We write something like:

> install time is going to increate and may take 45 or more minutes alone to complete seeding the collections.

downstream/modules/topologies/ref-cont-a-env-a.adoc
elyezer, line 14: Is you like the previous suggestion, then we would need to also reflect it here

elyezer, line 14: Is you like the previous suggestion, then we would need to also reflect it here

ariordan-redhat (approved):
LGTM

PR #2844, opened 2025-01-30
2.5 Update CONT-A to include RAM requirement for bundled install (#2840)

michellemacrh:
* Update CONT-A to include RAM requirement for bundled install

- Add a note to Tested deployment models doc to state that 32 GB RAM is required for a bundled install for CONT-A
- Add the same note to Containerized installation doc
- Remove the previous warning that was in place as it's no longer relevant now we have the required RAM config
- Archive the warning snippet

Update Docs for hub_seed_collections=true for containerized installer

https://issues.redhat.com/browse/AAP-38830

* Updates based on technical review


ianf77 (approved):
LGTM

PR #2849, opened 2025-01-31
Adding missing tab info to 2.x version

g-murray:
[AAP-39490](https://issues.redhat.com/browse/AAP-39490) requests adding some information on the tags tab as part of the 2.5 release but the same information is relevant for the current version also.


EMcWhinn (approved):
Small suggestion left.

PR #2867, opened 2025-02-04
Updating awx to AutomationController

g-murray:
[AAP-39765](https://issues.redhat.com/browse/AAP-39765)

Awx references in the Installing on OCP guide and Backup and restore guide that need to be updated to AutomationController

Edits made but want confirmation before publishing [See Slack](https://redhat-internal.slack.com/archives/C019LBHATBP/p1738334503825599)


ariordan-redhat (approved):
LGTM

PR #2876, opened 2025-02-05
Removing Lightspeed from code block

g-murray:
[AAP-39552](https://issues.redhat.com/browse/AAP-39552)

AAP install fails when lightspeed disabled false is in yaml - removing from code block


ariordan-redhat (approved):
LGTM

PR #2902, opened 2025-02-10
Add restore procedure to Containerized installation

michellemacrh:
Containerized installation - Add public documentation for restore from backup

https://issues.redhat.com/browse/AAP-34722

### DIFF
--- a/downstream/modules/platform/proc-restore-aap-container.adoc
+++ b/downstream/modules/platform/proc-restore-aap-container.adoc
@@ -1,0 +1,23 @@
+[id="proc-restore-aap-container"]
+
+= Restoring container-based {PlatformNameShort}
+
+Restore your {ContainerBase} of {PlatformNameShort} from a backup.
+
+.Procedure
+
+. Go to the {PlatformName} installation directory on your {RHEL} host.
+
+. Run the `restore` playbook:
++
+----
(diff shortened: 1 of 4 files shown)


downstream/modules/platform/proc-restore-aap-container.adoc
```diff
@@ -0,0 +1,23 @@
...
+
+. Go to the {PlatformName} installation directory on your {RHEL} host.
+
+. Run the `restore` playbook:
++
+----
+$ ansible-playbook -i <path_to_inventory> ansible.containerized_installer.restore
+----
+
+This will restore the important data deployed by the containerized installer such as:
```
g-murray, line 17: Try to avoid using the future tense, I think you'll get away with "This  restores" here.

downstream/modules/platform/ref-general-inventory-variables.adoc
```diff
@@ -25,7 +25,7 @@ RPM Default = `/var/lib/ansible-automation-platform-bundle`
 
 | `custom_ca_cert` | `custom_ca_cert` | The path to the custom CA certificate file.
 
-If set, this will install a custom CA certificate to the system trust store.
+If set, this will install a custom CA certificate to the system truststore.
```
g-murray, line 28: Same here "this installs"

g-murray (approved):
Two small suggestions, but still, LGTM.

PR #2905, opened 2025-02-10
Correct custom resource definition (#2904)

ianf77:
Set eda.disabled to true

BUG - Logic error within the custom resource definition for gateway that would deploy EDA instead of just connecting to Controller and Hub

https://issues.redhat.com/browse/AAP-40045


michellemacrh:
LGTM!

PR #2907, opened 2025-02-10
AAP-38726 - Removed incorrect access notes for org admin for tokens/apps

dcdacosta:
Updated the Access rules for applications and tokens to remove incorrect information related to organization administrators.


hherbly (approved):
LGTM

PR #2917, opened 2025-02-12
Adding CR details to Installing chapter

g-murray:
AnsibleLightspeed Custom Resource and Ansible Lightspeed Operator is being deployed when installing the AAP 2.4 and 2.5 operator which is not accounted for in the documentation

- Adding a verification section to the install chapter where a user can confirm what should be including in their deployment 
- Also touched up the information for cluster scoped channel


michellemacrh:
LGTM

PR #2925, opened 2025-02-12
AAP-39927 Added descriptions for the service level administrator checkboxes

dcdacosta:
This PR adds the service level administrator checkboxes and a description for Editing migrated admin users.


downstream/modules/platform/proc-gw-editing-a-user.adoc
```diff
@@ -6,7 +6,13 @@
...
+In upgrade scenarios, there might be pre-existing user accounts from {ControllerName} or {HubName} services. When editing these user accounts, the *User type* checkboxes indicate whether the account had one of the following service level administrator privileges: 
+
+Automation Execution Administrator:: A previously defined automation controller administrator with full read and write privileges over automation execution resources only.
+Automation Decisions Administrator:: A previously defined Event-Driven Ansible administrator with full read and write privileges over automation decision resources only.
+Automation Content Administrator:: A previously defined automation hub administrator with full read and write privileges over automation hub resources only.
```
hherbly, line 13: Consider using attributes in these three definitions? For example, 
A previously defined {ControllerName} administrator
A previously defined {HubName} administrator...etc

PR #2921, opened 2025-02-12
AAP-40078 updated important note

emurtoug:
[AAP-40078](https://issues.redhat.com/browse/AAP-40078) 

Files modified:
- assembly-network-ports-protocols.adoc

PR #2947, opened 2025-02-17
AAP-40078 updated important note (#2921)

emurtoug:
* AAP-40078 updated important note

* AAP-40078 updated important note

* AAP-40078 updated important note


michellemacrh:
LGTM

PR #3033, opened 2025-03-05
AAP-40672 Devtools - modify devcontainer creation

ariordan-redhat:
Update the procedure for scaffolding a devcontainer file.
Affects `titles/develop-automation-content`

### DIFF
--- a/downstream/modules/devtools/proc-devtools-install-container.adoc
+++ b/downstream/modules/devtools/proc-devtools-install-container.adoc
@@ -3,7 +3,7 @@
 = Installing {ToolsName} on a container inside {VSCode}
 
 The Dev Containers {VSCode} extension requires a `.devcontainer` file to store settings for your dev containers. 
-You must create a config file for your dev container and reopen your directory in a container in {VSCode}.
+You must use the Ansible extension to scaffold a config file for your dev container, and reopen your directory in a container in {VSCode}.
 
 .Prerequisites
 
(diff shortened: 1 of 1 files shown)


downstream/modules/devtools/proc-devtools-install-container.adoc
```diff
@@ -20,26 +20,21 @@ xref:devtools-setup-registry-redhat-io_installing-devtools[Authenticating with t
 
 .Procedure
 
-. In {VSCode}, open the directory where you want to store the configuration files for your development container.
-. Create a subdirectory called `.devcontainer`.
-. In the `.devcontainer` directory, create a file called `devcontainer.json`.
+. In {VSCode}, navigate to your project directory.
+. Click the Ansible icon in the {VSCode} activity bar to open the Ansible extension.
```
g-murray, line 24: Do you have an image of the icon by any chance? Not vital but would follow suit with how most other refer to icons

g-murray (approved):
One suggestion, but lGTM

downstream/modules/devtools/proc-devtools-install-container.adoc
ariordan-redhat: I was wondering about that. I don't have an image small enough to use inline, and it'd look a bit silly to have it on a line of its own - it's just the usual "A" icon.
I need to revamp these docs in the next week or so because the UI has been updated, so I'll add something about the icon then if that's OK - I can add af ull screenshot and point to it there or something.

PR #3049, opened 2025-03-10
Remove port from diagram and table (#3048)

ianf77:
Remove port 50051 from network diagram and ports table

https://issues.redhat.com/browse/AAP-41674


g-murray (approved):
LGTM

PR #3058, opened 2025-03-10
Update install proc for Containerized installation

michellemacrh:
Containerized installation - add install command for install with extra vars

https://issues.redhat.com/browse/AAP-40801

### DIFF
--- a/downstream/modules/platform/proc-installing-containerized-aap.adoc
+++ b/downstream/modules/platform/proc-installing-containerized-aap.adoc
@@ -41,7 +41,7 @@ ansible-playbook -i inventory -e @vault.yml --ask-vault-pass -K -v  ansible.cont
 * `-K` - (Optional) If your privilege escalation requires you to enter a password, add this to the installation command. You are then prompted for the BECOME password.
 * `-v` - (Optional) You can use increasing verbosity, up to 4 v’s (`-vvvv`) to see the details of the installation process. However, it is important to note that this can significantly increase installation time, so use it only as needed or requested by Red Hat support.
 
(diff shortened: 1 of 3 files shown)


downstream/modules/platform/proc-installing-containerized-aap.adoc
```diff
@@ -15,6 +26,24 @@ For example:
...
+----
+ansible-playbook -i inventory -e @vault.yml --ask-vault-pass -K -v  ansible.containerized_installer.install
+----
+
+* `-i <inventory_file_name>` - The inventory file to use for the installation.
+* `-e @<vault_file_name> --ask-vault-pass` - (Optional) If you are using a vault to store sensitive variables, add this to the installation command.
+* `-K` - (Optional) If your privilege escalation requires you to enter a password, add this to the installation command. You are then prompted for the BECOME password.
+* `-v` - (Optional) You can use increasing verbosity, up to 4 v’s (`-vvvv`) to see the details of the installation process. However, it is important to note that this can significantly increase installation time, so use it only as needed or requested by Red Hat support.
+
+Installation of containerized {PlatformNameShort} will begin.
```
g-murray, line 44: Optional: Try to avoid the use of future case. You could get away with  saying "The installation of......begins" here I think.

g-murray (approved):
LGTM

downstream/modules/platform/proc-installing-containerized-aap.adoc
michellemacrh: That's fair, I always think it sounds really funny "The installation begins..." but you're right 😆  will update!

PR #3070, opened 2025-03-11
Add callout for RHEL support in aap-docs

michellemacrh:
Minimum Supported Versions Need to Call Out Specific Major RHEL Versions

https://issues.redhat.com/browse/AAP-41627


ianf77 (approved):
LGTM

PR #3071, opened 2025-03-11
2.5 Add callout for RHEL support in aap-docs (#3070)

michellemacrh:
Minimum Supported Versions Need to Call Out Specific Major RHEL Versions

https://issues.redhat.com/browse/AAP-41627


hherbly (approved):
LGTM

PR #3090, opened 2025-03-14
AAP-42062 Correct the URL for the EDA API docs in Chapt 1 of Using automation decisions

jself-sudoku:
In the Notes admonition of [Chapter 1. Event-Driven Ansible controller overview](https://docs.redhat.com/en/documentation/red_hat_ansible_automation_platform/2.5/html-single/using_automation_decisions/index#eda-user-guide-overview), updated the current URL for EDA API docs

### DIFF
--- a/downstream/assemblies/eda/assembly-eda-user-guide-overview.adoc
+++ b/downstream/assemblies/eda/assembly-eda-user-guide-overview.adoc
@@ -24,7 +24,7 @@ The following procedures form the user configuration:
 [NOTE]
 
 ====
-* API documentation for {EDAcontroller} is available at \https://<platform-host>/api/eda/v1/docs
+* API documentation for {EDAcontroller} is available at \https://<gateway-host>/api/eda/v1/docs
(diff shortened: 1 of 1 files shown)


downstream/assemblies/eda/assembly-eda-user-guide-overview.adoc
```diff
@@ -24,7 +24,7 @@ The following procedures form the user configuration:
 [NOTE]
 
 ====
-* API documentation for {EDAcontroller} is available at \https://<eda-server-host>/api/eda/v1/docs
+* API documentation for {EDAcontroller} is available at \https://<platform-host>/api/eda/v1/docs
```
zkayyali812, line 27: Im not sure if instead of calling this `platform-host`, if we should instead call this `gateway-host`? 
Perhaps gateway host may be better, since this is more inline with terminology here - https://docs.redhat.com/en/documentation/red_hat_ansible_automation_platform/2.5/html/rpm_installation/assembly-platform-install-scenario#con-install-scenario-recommendations

rogrange: LGTM

PR #3097, opened 2025-03-18
AAP-42057 edits

g-murray:
[AAP-42057](https://issues.redhat.com/browse/AAP-42057)

We need to add some context around namespaces, in particular not using default as the namespace. SME approved messaging:

Deploying AAP in the default namespace on your openshift cluster is not allowed. The aap namespace is recommended. You can use a custom namespace, however AAP should be the only service running in that namespace.


michellemacrh:
LGTM

PR #3124, opened 2025-03-20
AAP-41494 - Add new Auto migrate users setting to auth procedures

dcdacosta:
This PR updates the authentication configuration procedures to include the new Auto migrate users from selection to the steps. Since the setting is the same for all types, I've created a snippet for this information.


EMcWhinn (approved):
LGTM

PR #3147, opened 2025-03-21
AAP-37305 Document how to copy an existing rulebook activation

jself-sudoku:
[AAP-37305](https://issues.redhat.com/browse/AAP-37305) Create a sub-section in the [Rulebook Activation chapter of the Using automation decisions guide](https://docs.redhat.com/en/documentation/red_hat_ansible_automation_platform/2.5/html-single/using_automation_decisions/index#eda-rulebook-activations) that focuses on how to create a copy of an existing rulebook activation.


bzwei (approved):

tiyiprh (approved):

downstream/modules/eda/proc-eda-copy-rulebook-activation.adoc
```diff
@@ -0,0 +1,33 @@
...
+====
+The original rulebook activation continues to run after you have copied it. If you try to enable the copied activation without editing the fields (including the Name field) to distinguish it from the original, a message is displayed reminding you that the rulebook activation was copied from an original, and enabling it might fail or result in duplicate jobs and other complications.
+====
+ 
+. Before you run the copied rulebook activation, edit the fields by completing the following: 
+.. Next to the copied rulebook activation, click the *Edit* icon. This takes you to the Edit form. 
+.. Edit the desired fields.
++
+[NOTE]
+====
```
sayjadha, line 27: Suggestion: Can this be a step in the procedure, perhaps after step 'select copy rulebook activation' and the following note where it's mentioned the rulebook activation is created with the same name as that of the original one? It might get missed in the note. Your call though.

sayjadha (commented):
One minor suggestion, rest LGTM.

downstream/modules/eda/proc-eda-copy-rulebook-activation.adoc
jself-sudoku: The goal is to keep the edit task as simple as possible since we are creating an entire topic on editing activations. That's why we just say edit the fields. Just allowing the user to edit fields they desire, but making sure they know the significance of editing the Name field. SMEs didn't want there to be focus on the Name field only.

sayjadha: Merging per @jself-sudoku 's request.

PR #3171, opened 2025-03-25
AAP-42638: removes info about analytics opt-in in first login process

hherbly:
This PR removes a step instructing the user to opt-in to automation analytics from the first login flow. See [AAP-42638](https://issues.redhat.com/browse/AAP-42638) for more.


jself-sudoku:
lgtm

PR #3173, opened 2025-03-26
Conditionalize assembly to include in two guides

ianf77:
Replace includes for Access management and conditionalize for Operations Guide

Conditionalize assembly for Access management

https://issues.redhat.com/browse/AAP-42657


dcdacosta (approved):
LGTM!

PR #3181, opened 2025-03-26
41494 - removed details about migrating via the API

dcdacosta:
Support for migrating legacy auth users is now supported in the UI. This PR removes the steps to manually configure this setting via the API and includes information about using the Auto migrate users from UI selection instead.


jself-sudoku:
lgtm

PR #3190, opened 2025-03-27
Update receptor inventory file vars

michellemacrh:
Receptor variables - Appendix Inventory file variables improvements

https://issues.redhat.com/browse/AAP-40603


downstream/modules/platform/ref-receptor-inventory-variables.adoc
```diff
@@ -3,82 +3,122 @@
...
+
+| 
+| `receptor_disable_signing` 
+| Controls whether signing of communications between receptor nodes is enabled or disabled. +
+Set this variable to `true` to disable communication signing.
+| Optional
+| `false`
+
+| 
+| `receptor_disable_tls`
```
dsavineau, line 69: RPM installer has the `receptor_tls` variable (default to true)

michellemacrh: Thanks @dsavineau I added `receptor_tls` on its own row; just because one is to enable and one is to disable 👍

dsavineau (approved):

PR #3192, opened 2025-03-27
Adding second part of Manage devices section

EMcWhinn:
[Docs] Edge Manager - Re-work Managing devices chapter

https://issues.redhat.com/browse/AAP-42675


downstream/assemblies/platform/assembly-edge-manager-install.adoc
```diff
@@ -4,3 +4,9 @@
 
 Enable the {RedHatEdge} to manage edge devices and applications at scale.
 
+ADD ENABLING/INSTALL FOR AAP CONTENT HERE WHEN AVAILABLE
```
g-murray, line 7: Should this be // commented out?

downstream/modules/platform/proc-edge-manager-config-git-cli.adoc
```diff
@@ -0,0 +1,71 @@
...
+apiVersion: flightctl.io/v1alpha1
+kind: Repository
+metadata:
+  name: site-settings
+spec:
+  type: git
+  url: https://github.com/<your_org>/<your_repo>.git
+----
+
+. Create the `Repository` resource by running the following command:
```
g-murray, line 25: Totally optional here, but you could stop at running, "the following command" isn't really necessary. 

There are a few of these, in this PR. As its optional I'm not going to call them all out.

downstream/modules/platform/ref-edge-manager-rule-files.adoc
```diff
@@ -0,0 +1,85 @@
...
+|`EnvVars` | Optional. A list of key-value pairs to set as environment variables for the command.
+
+|`WorkDir` | Optional. The directory the command is run from.
+
+|`Timeout` | Optional. The maximum duration that is allowed for the action to complete. Specify the duration as a single positive integer followed by a time unit. The `s`, `m`, and `h` units are supported for seconds, minutes, and hours.
+
+|`If` | Optional. A list of conditions that must be true for the action to be run. If not provided, actions run unconditionally.
+|===
+
+By default, actions are performed every time the hook is triggered.
```
g-murray, line 41: Active voice : "By default, the system performs actions every time the hook is triggered."

downstream/modules/platform/ref-edge-manager-rule-files.adoc
```diff
@@ -0,0 +1,85 @@
...
+By default, actions are performed every time the hook is triggered. 
+However, for the `afterUpdating` hook, you can use the `If` parameter to add conditions that must be true for an action to be performed. 
+Otherwise, the action is skipped. 
+
+For example, to run an action only if a given file or directory changes during the update, you can define a path condition that takes the following parameters:
+
+|===
+|Parameter | Description
+|`Path` | An absolute path to a file or directory that must change during the update as condition for the action to be performed.
+Specify paths by using forward slashes (`/`). If the path is to a directory, it must end with a forward slash (`/`).
```
g-murray, line 50: Just a suggestion, but I think turning all these "If" statements into bullets would help with readability. It's a fairly dense block as is.

g-murray (commented):
Few suggestions some optional, not the first one though!

downstream/assemblies/platform/assembly-edge-manager-install.adoc
EMcWhinn: Sorry I should of clarified, this is intentional. Doc isn't on the portal yet and I want to highlight in any drafts that this has not yet been supplied as I've asked numerous times.

downstream/modules/platform/proc-edge-manager-config-git-cli.adoc
EMcWhinn: Yes I see what you mean, it was brought in from the ACM peer reviews so its consistent through the whole doc. I'll keep it for now but might be better to remove before publishing.

g-murray (approved):
LGTM

PR #3204, opened 2025-04-01
AAP-37304 Document how to edit an existing activation

jself-sudoku:
Per ticket AAP-37304, a new topic has been added to the [Rulebook activations chapter in the Using automation decisions guide](https://docs.redhat.com/en/documentation/red_hat_ansible_automation_platform/2.4/html/event-driven_ansible_controller_user_guide/eda-rulebook-activations): Editing a rulebook activation.


downstream/modules/eda/proc-eda-edit-rulebook-activation.adoc
```diff
@@ -0,0 +1,32 @@
...
+
+. On the Rulebook Activations page, next to the activation you want to edit, toggle the btn:[Rulebook Activation enabled] button to the off position first to disable the activation.
++
+The *Disable rulebook activations* message is displayed asking you to confirm that you want to disable the activation.
+. Select the *Yes, I confirm that I want to disable these <1> rulebook activations* checkbox and click btn:[Disable rulebook activations]. 
+. Next to the rulebook activation, click the *Edit* icon. This takes you to the Edit form.
++
+[NOTE]
+====
+You can also access the *Edit* feature by clicking the rulebook activation on the Rulebook Activations page, toggling the btn:[Rulebook activation enabled] button to the off position, confirming that you want to disable the activation, and clicking the btn:[Edit rulebook activation] button on the top right of the page to access the Edit form.
```
bzwei, line 22: Shouldn't it be Rulebook Activation Details page?

jself-sudoku: We refer to the page that lists all of the rulebook activations as Rulebook Activations page throughout our content. Why would we be inconsistent here?

B-Whitt (approved):
lgtm!

jself-sudoku:
Thanks for the approval, @B-Whitt !

downstream/modules/eda/proc-eda-edit-rulebook-activation.adoc
bzwei: Yes. There are two pages in the process. First you click the activation on the Rulebook Activaitons page. It lands at the details page. Then on the details page you toggle the Rulebook activation enabled button. When I read the text it seems to me there is only one page, and the toggling button is on the first page. If I don't read carefully this process is just like the same as one above.

tiyiprh: What @bzwei said is correct. On step 1 it's correct to call it the Rulebook Activations page (because that is the table view listing all the activations). On step 5, it should technically be the Details page vs the table view. 
I think if we want to make the note clearer: We can may say something like:
"You can also access the Edit feature by clicking the rulebook activation on the Rulebook Activations page, which takes you to the details page of the rulebook activation..." or something like that.

jself-sudoku: In this instance, the user still has to start on the Rulebook Activations page. That's the reference here. They don't get to the Details page until they click the activation, which is found on the Rulebook Activations page. I was trying to summarize here because technically, we aren't supposed to add procedures to an admonition (Notes, Important, etc.). The guidance from our RH supplementary guide states: "Admonitions should be short and concise. Do not include procedures in an admonition." Also, this is thinking beyond the Rulebook Activations page.

bzwei (approved):

downstream/modules/eda/proc-eda-edit-rulebook-activation.adoc
```diff
@@ -0,0 +1,32 @@
...
+You can edit a rulebook activation after you have created or run it to correct input for fields (log levels, Restart policy, turn auditing off or on, and the like) or  help mitigate issues caused by failure.
+
+.Procedure
+
+[NOTE]
+====
+Before you can edit any activation, you must disable it first.
+====
+
+. On the Rulebook Activations page, next to the activation you want to edit, toggle the btn:[Rulebook Activation enabled] button to the off position first to disable the activation.
```
hherbly, line 14: This first step repeats what's in the note directly above it--is this deliberate?

jself-sudoku: Yes, I asked about that and one of the SMEs that it was important to emphasize the importance. I agree with you that it's a bit redundant. The way things are set up you can't edit unless you disable the activation. Just wanted a second opinion from a docs perspective. I'm going to remove it for now. Thanks for the input, @hherbly !

hherbly (approved):
LGTM!

PR #3215, opened 2025-04-04
Fixing level offset issue

g-murray:
Level offset was incorrectly set to +2 instead of +1. This PR fixes that.


michellemacrh:
LGTM!

PR #3235, opened 2025-04-09
Editing container group proc in 2.4 doc

EMcWhinn:
Container Group documentation is incorrect for recent versions of OpenShift

https://issues.redhat.com/browse/AAP-41641


michellemacrh:
LGTM!

PR #3241, opened 2025-04-09
AAP-39723: updates login flow to push EEs to PAH

hherbly:
Work tracked in [AAP-39723](https://issues.redhat.com/browse/AAP-39723).


EMcWhinn (approved):
LGTM

PR #3246, opened 2025-04-10
AAP-43744 updating title for 2.4 and 2.5

g-murray:
[AAP-43744](https://issues.redhat.com/browse/AAP-43744) Update the wording for section 4.2

The wording in section 4.2 for the 2.5 AAP Operator docs is confusing.  I think we should refer to "Installing the AAP Operator in a namespace" as that, rather than "Subscribing a namespace to an operator".


michellemacrh:
LGTM!

PR #3291, opened 2025-04-16
Update Using custom TLS certificates in Containerized installation

michellemacrh:
[AAP Setup] gateway_tls_cert and gateway_tls_key is not working as expected.

https://issues.redhat.com/browse/AAP-39956


jonquilwilliams:
LGTM

PR #3292, opened 2025-04-16
Update python version

ianf77:
AAP 2.5 EE uses Python 3.11 introducing non-obvious manual changes

https://issues.redhat.com/browse/AAP-40730


michellemacrh:
LGTM!

PR #2978, opened 2025-02-20
Adding service account info to controller docs

EMcWhinn:
Document controller changes needed for service accounts

https://issues.redhat.com/browse/AAP-36066

Affects `titles/controller-user-guide``

PR #3296, opened 2025-04-16
May service account updates

EMcWhinn:
Update Configuring Automation Execution doc to reflect service account changes to subscription management

https://issues.redhat.com/browse/AAP-42976

### DIFF
--- a/downstream/modules/platform/proc-controller-configure-analytics.adoc
+++ b/downstream/modules/platform/proc-controller-configure-analytics.adoc
@@ -7,19 +7,19 @@ When you imported your license for the first time, you were automatically opted 
 .Prerequisites
 
 * A service account created with the Analytics Administrator viewer role in console.redhat.com.
-For more information see, link:https://docs.redhat.com/en/documentation/red_hat_hybrid_cloud_console/1-latest/html/creating_and_managing_service_accounts/proc-ciam-svc-acct-overview-creating-service-acct#proc-ciam-svc-acct-create-creating-service-acct[Creating a service account].
+For more information, see link:https://docs.redhat.com/en/documentation/red_hat_hybrid_cloud_console/1-latest/html/creating_and_managing_service_accounts/proc-ciam-svc-acct-overview-creating-service-acct#proc-ciam-svc-acct-create-creating-service-acct[Creating a service account].
 
 .Procedure
 . From the navigation panel, select {MenuSetSubscription}.
-The *Subscription* page is displayed.
+The *Subscription Settings* page is displayed.
 . If you have not already set up a subscription, do so now, and ensure that on the next page you have selected *{Analytics}* to use analytics data to enhance future releases of {PlatformNameShort} and to give the Red Hat Insights service to subscribers.
-+
-image::automation_analytics.png[Automation analytics page]
+//+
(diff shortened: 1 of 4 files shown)


downstream/modules/platform/proc-controller-find-subscription.adoc
```diff
@@ -0,0 +1,15 @@
...
+
+= Finding your subscription
+
+Find your subscription after installation.
+
+.Procedure
+
+. From the navigation panel, select {MenuSetSubscription}.
+The *Subscription Settings* page is displayed.
+. Select the *Username / password* option.
```
hherbly, line 11: I think this is still being decided, but I think the username/password tab will be changed to client ID/client secret. There's a feature lead sync meeting for service accounts tomorrow right after our team meeting--I'll try to confirm whether or not this change is happening and I'll let you know.

downstream/modules/platform/proc-controller-configure-analytics.adoc
```diff
@@ -6,30 +6,21 @@ When you imported your license for the first time, you were automatically opted
 
 .Prerequisites
 
-* A service account created with the Analytics Administrator viewer role in console.redhat.com.
-For more information see, link:https://docs.redhat.com/en/documentation/red_hat_hybrid_cloud_console/1-latest/html/creating_and_managing_service_accounts/proc-ciam-svc-acct-overview-creating-service-acct#proc-ciam-svc-acct-create-creating-service-acct[Creating a service account].
+* A service account created with the *Automation Analytics Viewer* role in console.redhat.com.
```
hherbly, line 9: I haven't heard anything about adding a role for updating the Analytics credentials. Is this legit? If so, I may need to update my own procedure, but again I'm not familiar with the requirement. I know you need to add the service account to a user group for the purpose of finding your subscription...I could shoot this over to Seth to see what he says. Lmk if you'd like me to do that

downstream/modules/platform/proc-controller-find-subscription.adoc
```diff
@@ -0,0 +1,38 @@
...
+.. In the *Client ID / Satellite username* field enter the client ID you received when you created your service account.
+.. In the *Client secret / Satellite password* field enter the client secret you received when you created your service account.
+Your subscription appears in the list menu labeled *Subscription*. 
+Select your subscription.
+
+. After you have added your subscription, click btn:[Next].
+. Check the box indicating that you agree to the *End User License Agreement*.
+. Review your information and click *Finish*.
+
+.Troubleshooting
```
hherbly, line 22: I'm not sure about adding a troubleshooting section to the documentation. I chatted with Donna and Lynne about it, and they agreed that it would be better to just link to the KB article instead of including the troubleshooting steps in the doc set (especially since it's kind of a doc-around). I'll slack you our respective PRs so you can see how we both did it.

downstream/modules/platform/proc-controller-configure-analytics.adoc
EMcWhinn: This comes from Seth's comment back in February for the original service account updates, so I'm not sure if things have changed since then?

https://github.com/ansible/aap-docs/pull/2978/files#r1974082472

PR #3299, opened 2025-04-17
AAP-42905 - Corrected gateway token API path

dcdacosta:
This PR updates the gateway api token path as described in https://issues.redhat.com/browse/AAP-42905.


EMcWhinn (approved):
LGTM

PR #3308, opened 2025-04-22
AAP-44196 Adding Tip Admonition

g-murray:
[AAP-44196](https://issues.redhat.com/browse/AAP-44196) - Add clarity on name variable usage in CRs

Adding the following tip to help users understand the 'name' variable has multi use

"Tip: You can link in existing components by specifying the component name under the `name` variable. You can also use `name` to create a custom name for a new component."

### DIFF
--- a/downstream/assemblies/platform/assembly-appendix-operator-crs.adoc
+++ b/downstream/assemblies/platform/assembly-appendix-operator-crs.adoc
@@ -9,6 +9,12 @@ ifdef::context[:parent-context: {context}]
 
 This appendix provides a reference for the {PlatformNameShort} custom resources for various deployment scenarios.
 
+[TIP]
+====
+You can link in existing components by specifying the component name under the `name` variable.
+You can also use `name` to create a custom name for a new component.
+====
+
 include::platform/ref-operator-crs.adoc[leveloffset=+1]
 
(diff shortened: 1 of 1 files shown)


downstream/assemblies/platform/assembly-appendix-operator-crs.adoc
```diff
@@ -9,6 +9,12 @@ ifdef::context[:parent-context: {context}]
 
 This appendix provides a reference for the {PlatformNameShort} custom resources for various deployment scenarios.
 
+[TIP]
+====
+You can link in existing components by specifying the component name under the `name` variable.
```
ianf77, line 14: Do you link in i.e pull them in somehow, or link to: provide a link to?

ianf77 (commented):
One question.

downstream/assemblies/platform/assembly-appendix-operator-crs.adoc
g-murray: By using the same name as an existing deployment in your namespace, the AAP Operator recognizes and uses that deployment - so it links or pulls it in.  Rather than say creating a new deployment with that name. If the name doesn't match anything that already exists it will just create a new deployment. That make sense?

PR #3321, opened 2025-04-24
Add attribute for Installer

ariordan-redhat:
Adds an attribute for the installer / installation program, to minimise the impact of future name changes 

.

PR #3323, opened 2025-04-24
2.5 Add attribute for Installer (#3321)

ariordan-redhat:
Adds an attribute for the installer / installation program, to minimise the impact of future name changes


hherbly (approved):
LGTM

PR #3342, opened 2025-04-28
Update Adding execution nodes in Containerized installation

michellemacrh:
Update ref-adding-execution-nodes.adoc to denote that using groups for receptor_peers is not currently possible.

Containerised Installation (Mesh) - Can't establish connections from the execution nodes to the controller nodes

https://issues.redhat.com/browse/AAP-42614


g-murray (approved):
LGTM

PR #3358, opened 2025-04-29
AAP-43858 AAP-43859 Fix YAML syntax in RHDH operator install

ariordan-redhat:
AAP-43858 AAP-43859 Fix YAML syntax in RHDH operator install


michellemacrh:
LGTM!

PR #3359, opened 2025-04-29
2.5 AAP-43858 AAP-43859 Fix YAML syntax in RHDH operator install (#3358)

ariordan-redhat:
Backports #3358 
AAP-43858 AAP-43859 Fix YAML syntax in RHDH operator install and uninstall procedures
Affects `/titles/aap-plugin-rhdh-install`


michellemacrh:
LGTM

PR #3372, opened 2025-05-01
Update tested topology support wording

michellemacrh:
The topology support statement in Tested deployment models is updated to clarify support for deployments outside of the published reference architectures.

Update Statement for Supported Topologies

https://issues.redhat.com/browse/AAP-44586


g-murray (approved):
LGTM

PR #3381, opened 2025-05-02
Add compression vars to backup proc for Containerized installation

michellemacrh:
Docs - Containerized installation - Document use of compression when performing a backup

https://issues.redhat.com/browse/AAP-39129

### DIFF
--- a/downstream/modules/platform/proc-backup-aap-container.adoc
+++ b/downstream/modules/platform/proc-backup-aap-container.adoc
@@ -8,6 +8,37 @@ Perform a backup of your {ContainerBase} of {PlatformNameShort}.
 
 . Go to the {PlatformName} installation directory on your {RHEL} host.
 
+. To control compression of the backup artifacts before they are sent to the host running the backup operation you can use the following variables in your inventory file:
+.. For control of compression for filesystem related backup files:
++
+----
+# For global control of compression for filesystem related backup files 
+use_archive_compression=true
+
+# For component-level control of compression for filesystem related backup files
+#controller_use_archive_compression=true
+#eda_use_archive_compression=true
+#gateway_use_archive_compression=true 
+#hub_use_archive_compression=true
+#pcp_use_archive_compression=true
+#postgresql_use_archive_compression=true
+#receptor_use_archive_compression=true
+#redis_use_archive_compression=true
+----
++
+.. For control of compression for database related backup files:
++
+----
+# For global control of compression for database related backup files 
+use_db_compression=true  
+
+# For component-level control of compression for database related backup files
+#controller_use_db_compression=true
+#eda_use_db_compression=true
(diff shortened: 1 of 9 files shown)


downstream/modules/platform/proc-backup-aap-container.adoc
```diff
@@ -8,6 +8,37 @@ Perform a backup of your {ContainerBase} of {PlatformNameShort}.
 
 . Go to the {PlatformName} installation directory on your {RHEL} host.
 
+. To control compression of the backup artifacts before they are sent to the host running the backup operation you can use the following variables in your inventory file:
```
EMcWhinn, line 11:
```suggestion
. To control compression of the backup artifacts before they are sent to the host running the backup operation, you can use the following variables in your inventory file:
```

downstream/modules/platform/ref-gateway-variables.adoc
```diff
@@ -223,12 +217,30 @@
...
+
 | `automationgateway_user_headers`
 | `gateway_nginx_user_headers`
 | List of additional NGINX headers to add to {Gateway}'s NGINX configuration.
 | Optional
 | `[]`
 
+| `automationgateway_verify_ssl` 
+|  
+| Denotes whether or not to verify {Gateway}'s web certificates when making calls from {Gateway} to itself during installation. Set to false to disable web certificate verification.
```
EMcWhinn, line 240:
```suggestion
| Denotes whether or not to verify {Gateway}'s web certificates when making calls from {Gateway} to itself during installation. Set to `false` to disable web certificate verification. 
```

downstream/modules/platform/ref-general-inventory-variables.adoc
```diff
@@ -187,6 +187,22 @@ For further information, see link:https://docs.ansible.com/ansible/latest/invent
 | Optional 
 | 
 
+| `use_archive_compression` 
+| `use_archive_compression`
+a| Controls at a global level whether the filesystem-related backup files are compressed before being sent to the host to run the backup operation. If set to `true`, a `tar.gz` file is generated on each {PlatformNameShort} host and then  gzip compression is used. If set to `false`, a simple tar file is generated.
```
EMcWhinn, line 192:
```suggestion
a| Controls at a global level whether the filesystem-related backup files are compressed before being sent to the host to run the backup operation. If set to `true`, a `tar.gz` file is generated on each {PlatformNameShort} host and then gzip compression is used. If set to `false`, a simple tar file is generated. 
```

EMcWhinn (commented):
Minor comments otherwise LGTM

downstream/modules/platform/ref-gateway-variables.adoc
michellemacrh: Thanks for catching this! will update

PR #3418, opened 2025-05-07
AAP-45308 Add sync frequency section to portal install doc

ariordan-redhat:
AAP-45308 Add sync frequency section to portal install doc


g-murray (approved):
LGTM

PR #3445, opened 2025-05-12
Self-service: Rename; add proc to disable telemetry

ariordan-redhat:
Add a procedure to disable telemetry data collection in Ansible self-service.


g-murray (approved):
LGTM

PR #3452, opened 2025-05-13
AAP-45780 Remove outdated modules

g-murray:
Removing recovery modules that are no longer needed as everything is through gateway now

Removing:

include::platform/proc-aap-controller-restore.adoc[leveloffset=+1]

include::platform/proc-aap-controller-yaml-restore.adoc[leveloffset=+1]

include::platform/proc-aap-hub-restore.adoc[leveloffset=+1]


michellemacrh:
LGTM!

PR #3453, opened 2025-05-13
Removing recovery modules that are no longer needed as everything is …

g-murray:
…through gateway now (#3452)


michellemacrh:
LGTM!

PR #3448, opened 2025-05-12
Add job template variables to using automation execution

EMcWhinn:
As an AAP User I want to know what variables it sets

https://issues.redhat.com/browse/AAP-45527

Affects `titles/controller-user-guide`

PR #3456, opened 2025-05-13
Add job template variables to using automation execution (#3448)

EMcWhinn:
* Add job template variables to using automation execution

As an AAP User I want to know what variables it sets

https://issues.redhat.com/browse/AAP-45527

* Changing location of variables ref file


michellemacrh:
LGTM

PR #3463, opened 2025-05-14
Correct typo in Creating and consuming execution environments

michellemacrh:
In con-additional-build-files.adoc the var `additional_build_steps` was referenced when it should be `additional_build_files`.

docs execution environment config has wrong value of additional_build_files

https://issues.redhat.com/browse/AAP-45864


g-murray (approved):
LGTM

PR #3465, opened 2025-05-14
2.5 Self-service: Rename; add proc to disable telemetry (#3445)

ariordan-redhat:
Rename AAP self-service technical preview to AAP self-service technology preview.
Add a procedure to disable telemetry.


g-murray (approved):
LGTM

PR #3539, opened 2025-05-27
Modularizing chapter 11 of controller user guide

EMcWhinn:
Using automation execution UI and modular compliance chapter 10

https://issues.redhat.com/browse/AAP-45087

Affects titles/controller-user-guide

### DIFF
--- a/downstream/modules/platform/proc-scm-insights.adoc
+++ b/downstream/modules/platform/proc-scm-insights.adoc
@@ -2,7 +2,12 @@
 
 = SCM Type - Configuring playbooks to use Red Hat Insights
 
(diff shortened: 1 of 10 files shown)


downstream/modules/platform/proc-scm-insights.adoc
```diff
@@ -2,7 +2,12 @@
 
 = SCM Type - Configuring playbooks to use Red Hat Insights
 
+Configure your projects to retrieve Ansible playbooks directly from Red Hat Insights. 
+By integrating with Red Hat Insights, you can use it to manage and deploy remediation playbooks identified through its analysis of your [RHEL] environment.
```
ianf77, line 6: Is RHEL the  attribute? Brackets are wrong.

PR #3550, opened 2025-05-27
AAP-44692 - Add details for enabling/disabling local auth

dcdacosta:
This PR adds details for enabling/disabling auth, the potential impacts and troubleshooting details.


hherbly (approved):
LGTM

PR #3558, opened 2025-05-28
replacing all OCP xrefs with links

g-murray:
As part of migration prep, replacing all xrefs with links.


michellemacrh (commented):
I think some of the links might be incorrect here. Building the docs locally resulted in some 404 errors - would recommend building the docs locally and testing to be sure :)

PR #3583, opened 2025-05-29
Removing incorrect formatting from controller UG (#3582)

EMcWhinn:
Removing incorrect use of ":_mod-docs-content-type:"

https://issues.redhat.com/browse/AAP-46762


g-murray (approved):
LGTM

PR #3593, opened 2025-05-30
AAP-39301: Adds domain of interest info to getting started with AAP guide

hherbly:
This work is being tracked in [AAP-39301](https://issues.redhat.com/browse/AAP-39301).

### DIFF
--- a/downstream/modules/platform/proc-set-domain-of-interest.adoc
+++ b/downstream/modules/platform/proc-set-domain-of-interest.adoc
@@ -1,0 +1,9 @@
+:_newdoc-version: 2.18.4
+:_template-generated: 2025-06-04
+:_mod-docs-content-type: PROCEDURE
+
+[id="set-domain-of-interest_{context}"]
+= Set your domains of interest
+
+After you have familiarized yourself with the dashboard, you can set your dashboard view so that features that you select are foregrounded, making them easy to see and interact with.
+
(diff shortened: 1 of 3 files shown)


downstream/modules/platform/proc-set-domain-of-interest.adoc
```diff
@@ -0,0 +1,23 @@
...
+. From the navigation panel, select {MenuAEJobs} or {MenuAETemplates}.
+. Beneath the page heading, next to *Domains*, is a list of topic-related labels. Select a label to filter jobs and job templates so that only content related to the labels is shown. You can choose more than one label. 
+. To clear your selection, click the *X*. 
+. To customize your domain options, select the wrench icon. In the modal that appears, select *Add Domain* to add new domains to filter with. 
+
+[NOTE]
+
+====
+
+You can add labels to your individual job templates by navigating to {MenuAETemplates}, selecting your job template, and clicking btn:[Edit template]. On the editing screen, enter the label you want to use in the *Labels* field and click btn:[Save job template]. Now the template will be one of the resources that appear when you filter with that domain label.
```
jonquilwilliams, line 21: This is totally optional, but if you want to use present tense for the last sentence, you could say "The template becomes one of the resources that appears when you filter with that domain label."

You could also make the last sentence as the intro:

"You can add labels to your individual job templates to make the templates appear as resources when you filter with that domain label. Navigate to..."

I think that it's fine as is, but one of these options might make it a bit more direct. Lmk if you prefer that I merge now.

PR #3618, opened 2025-06-09
AAP-45856 updated metrics-utility content

emurtoug:
[AAP-45856](https://issues.redhat.com/browse/AAP-45856) [Doc] Update published subscription and metrics-utility content in configuring automation execution

Files modified: 
- proc-configuring-the-metrics-utility.adoc

Title:
- controller-admin-guide

### DIFF
--- a/downstream/modules/platform/proc-configuring-the-metrics-utility.adoc
+++ b/downstream/modules/platform/proc-configuring-the-metrics-utility.adoc
@@ -5,7 +5,9 @@
 [id="configuring-the-metrics-utility"]
 = Configuring metrics-utility
 
-== On {RHEL} 
+Configuring `metrics-utility` on {RHEL} or {OCPShort} enables you to gather usage metrics from your {PlatformNameShort} and create various usage reports.
+
+== Configuring metrics-utility on {RHEL} 
 
 .Prerequisites:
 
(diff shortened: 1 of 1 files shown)


downstream/modules/platform/proc-configuring-the-metrics-utility.adoc
```diff
@@ -47,13 +56,24 @@ export METRICS_UTILITY_REPORT_EMAIL="email@email.com"
...
+----
++
+. To open the cron file for editing, run:
++
+[source, ]
+----
+crontab -e
 ----
 +
 . To configure the run schedule, add the following parameters to the end of the file and specify how often you want `metrics-utility` to gather information and build a report using link:https://www.redhat.com/sysadmin/linux-cron-command[cron syntax]. In the following example, the `gather` command is configured to run every hour at 00 minutes. The `build_report` command is configured to run every second day of each month at 4:00 AM.
```
himdel, line 79: lines 83/84 still mention 

```
0 */1 * * * metrics-utility gather_automation_controller_billing_data --ship --until=10m
0 4 2 * *  metrics-utility build_report
```

should be

```
0 */1 * * * /home/my-user/cron-gather
0 4 2 * * /home/my-user/cron-report
```

downstream/modules/platform/proc-configuring-the-metrics-utility.adoc
```diff
@@ -47,13 +56,24 @@ export METRICS_UTILITY_REPORT_EMAIL="email@email.com"
...
 ----
 +
-. Run the following command to gather and store the data in the provided `SHIP_PATH` directory:
+. To ensure that these files are executable, run: 
 +
 [source, ]
 ----
-metrics-utility gather_automation_controller_billing_data --ship --until=10m
+chmod a+x /home/my-user/cron-gather 
+/home/my-user/cron-report
```
himdel, line 69:
```suggestion
chmod a+x /home/my-user/cron-gather /home/my-user/cron-report
```

should be one line, not two, otherwise we're running the script, not making it runnable

downstream/modules/platform/proc-configuring-the-metrics-utility.adoc
```diff
@@ -100,22 +120,24 @@ May  8 09:46:26 ip-10-0-6-23 crontab[51659]: (root) END EDIT (root)
 metrics-utility build_report
 ----
```
himdel, line 121: This doesnt belong here?

If this section is about setting up a cron job, we probably don't want everyone to build a report while setting the cronjob?

(This may still make sense in a "after gather happened, you can try this to test if things will work ... but unless the user waits a month, build_report will fail because there would be no gathered data for the previous month.)

himdel: Those are the ones, yes, 115-122.

Because you wouldn't want to build a report when just trying to configure things, right?

Unless it's to check that things are configured right and that something was collected to build the report from .. but then you probably want to do the checking right away, so the command would be different because it'd hhave to run for "this month" instead of the default "last month" (and you'd need all the env variables that we've just hidden in the script).

(If we wanted to do that, then i'd suggest changing line 61 from `metrics-utility build_report` to `metrics-utility build_report "$@"`, to accept extra parameters when given, and then we'd change line 120 from `metrics-utility build_report` to something like `/home/my-user/cron-build --since=2d`, to make sure it picks up the freshly gathered data.)

emurtoug: Ok! I have commented out this step for now with a summary of our reasoning. If someone wants to add a build report example down the line, it should be easy to do. Thanks @himdel !

emurtoug:
Hi @himdel - this PR was replaced and added on to by https://github.com/ansible/aap-docs/pull/3660 I have closed AAP-45856, as these changes were made and merged in June 2025 (see Jira comments). My apologies for leaving both this PR and the Jira open!

himdel:
Ah, thanks! No worries, I missed that PR :)

PR #3619, opened 2025-06-09
Premigration tasks for Automation mesh

ianf77:
Split procedure

DITA pre-migration tasks. Automation mesh

https://issues.redhat.com/browse/AAP-47319

PR #3630, opened 2025-06-10
Premigration tasks for Automation mesh (#3619)

ianf77:
Split procedure

DITA pre-migration tasks. Automation mesh

https://issues.redhat.com/browse/AAP-47319


downstream/modules/platform/proc-deprovision-isolated-nodes.adoc
```diff
@@ -0,0 +1,22 @@
+[id="proc-deprovision-isolated-nodes"]
+
+= Deprovisioning isolated nodes
+You have the option to manually remove any isolated nodes using the `awx-manage` deprovisioning utility.
+
+[WARNING]
```
EMcWhinn, line 6: This admonition is not closed off with "====" and the procedure is included within it, is that intentional?

PR #3634, opened 2025-06-10
AAP on Azure DITA prep

g-murray:
[AAP-47163](https://issues.redhat.com/browse/AAP-47163) DITA pre-migration tasks: Ansible on Clouds

Fixed proc modules to adhere to DITA structure, and updated some procs that should be refs.


michellemacrh (approved):
LGTM!

PR #3646, opened 2025-06-11
Pre-migration Edge Manager updates

EMcWhinn:
DITA Edge Manager pre-migration tasks

https://issues.redhat.com/browse/AAP-47512

### DIFF
--- a/downstream/modules/platform/proc-edge-manager-build-image-QCoW2.adoc
+++ b/downstream/modules/platform/proc-edge-manager-build-image-QCoW2.adoc
@@ -1,0 +1,36 @@
+:_mod-docs-content-type: PROCEDURE
+
+[id="edge-manager-build-image-QCoW2"]
+
+= Building the QCoW2 disk image
+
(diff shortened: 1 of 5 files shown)


downstream/modules/platform/proc-edge-manager-build-image-QCoW2.adoc
```diff
@@ -0,0 +1,36 @@
+:_mod-docs-content-type: PROCEDURE
+
+[id="edge-manager-build-image-QCoW2"]
+
+= Building the QCoW2 disk image
+
+Red Hat OpenShift Virtualization can download disk images from an OCI registry but it expects a container disk image instead of an OCI artifact.
```
ariordan-redhat, line 7:
```suggestion
{OCPV} can download disk images from an OCI registry but it expects a container disk image instead of an OCI artifact. 
```

ariordan-redhat (commented):
One suggestion. Looks good to me otherwise

PR #3648, opened 2025-06-11
Update using custom TLS certs

michellemacrh:
Break the current module `ref-using-custom-tls-certificates.adoc` into appropriate modules. Additionally update the content based on the draft updates per the Jira.

Affected title: `titles/aap-containerized-install`

Document update for containerised AAP 2.5 enterprise topology custom certificate requirements

https://issues.redhat.com/browse/AAP-45145


downstream/assemblies/platform/assembly-using-custom-tls-certificates.adoc
```diff
@@ -0,0 +1,38 @@
...
+
+There are two primary ways to manage TLS certificates for your {PlatformNameShort} deployment: 
+
+. {PlatformNameShort} generated certificates (this is the default)
+. User-provided certificates
+
+// AAP generated certificates
+include::platform/con-installer-generated-certs.adoc[leveloffset=+1]
+
+== User-provided certificates
```
ianf77, line 18: Should this also be separate? A ref or a con?

ianf77 (commented):
One comment

downstream/assemblies/platform/assembly-using-custom-tls-certificates.adoc
michellemacrh: I wasn't sure if it was worth creating a separate module for what looks like an introductory sentence for the other modules 🤔 What do you think, is that worth doing?

PR #3676, opened 2025-06-16
Update ansible-core version reqs for containerized AAP

michellemacrh:
Add clarity about the ansible-core version requirements in Containerized installation.

Affected title: `titles/aap-containerized-install`

[Docs]Documentation gives mixed messages about ansible-core version for install

https://issues.redhat.com/browse/AAP-47449

### DIFF
--- a/downstream/snippets/cont-tested-system-config.adoc
+++ b/downstream/snippets/cont-tested-system-config.adoc
@@ -17,8 +17,12 @@ a|
 |
 
 | `ansible-core` 
-| `ansible-core` version {CoreUseVers} or later. 
-| {PlatformNameShort} uses the system-wide `ansible-core` package to install the platform, but uses `ansible-core` {CoreUseVers} for both its control plane and built-in execution environments. The required version is bundled within the platform's components, so you do not need to manually install it on your host system.
+a| 
+* For the installation: `ansible-core` version {CoreInstVers}. 
+* For {PlatformNameShort} operation: `ansible-core` version {CoreUseVers}.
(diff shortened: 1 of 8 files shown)


downstream/snippets/cont-tested-system-config.adoc
```diff
@@ -1,14 +1,31 @@
...
+| {RHEL} 9.2 or later minor versions of {RHEL} 9
+| 
+
+| CPU architecture 
+| x86_64, AArch64, s390x (IBM Z), ppc64le (IBM Power)
+|
+
+| `ansible-core` 
+| `ansible-core` version {CoreUseVers} or later. 
+| {PlatformNameShort} uses the system-wide `ansible-core` package to install the platform, but uses `ansible-core` {CoreUseVers} for both its control plane and built-in execution environments. The required version is bundled within the platform's components, so you do not need to manually install it on your host system.
```
dsavineau, line 21: Not sure to fully understand which ansible-core version and usage we're talking about so I'll try to explain to see if we're on the same page.

The containerized installer should use ansible-core 2.14 RPM package from RHEL 9 appstream repository (because we don't need to have AAP subscription enabled on the installer node).

The one from the AAP 2.5 repository shouldn't be used by the installer. It's possible but not mandatory. So running the installer with either 2.14 or 2.16 is working. Other later versions aren't verified.

However, the platform deployed by the installer will only use ansible-core 2.16 from AAP.

hherbly (approved):
LGTM

PR #3683, opened 2025-06-16
AAP-47754: updates attribute for policy settings

hherbly:
This work is being tracked in [AAP-47754](https://issues.redhat.com/browse/AAP-47754).


rogrange:
LGTM

PR #3725, opened 2025-06-25
Add controller and database descriptions

ianf77:
Missing components (Automation Controller and Database) in the Planning your installation document

https://issues.redhat.com/browse/AAP-48022

### DIFF
--- a/downstream/modules/platform/con-about-controller.adoc
+++ b/downstream/modules/platform/con-about-controller.adoc
@@ -2,4 +2,4 @@
 
 = {ControllerNameStart}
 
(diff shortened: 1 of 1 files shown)


downstream/modules/platform/con-about-controller.adoc
```diff
@@ -0,0 +1,5 @@
+[id="con-about-controller"]
+
+= {ControllerNameStart}
+
+An enterprise framework enabling users to define, operate, scale, and delegate Ansible automation across their enterprise.
```
ariordan-redhat, line 5:
```suggestion
{ControllerNameStart} is an enterprise framework enabling users to define, operate, scale, and delegate Ansible automation across their enterprise.
```

PR #3796, opened 2025-07-08
AAP-47431 Devtools pre-migration updates

ariordan-redhat:
Update the Devtools content to comply with dita rules for migration to the adobe tooling.


g-murray (approved):
LGTM

PR #3821, opened 2025-07-11
Content type editor updated

g-murray:
As part of our preparation for migrating to DITA I am using Jaromir's script to update the content type of all modules

https://github.com/jhradilek/content-type-editor

### DIFF
--- a/downstream/modules/platform/con-edge-manager-build-bootc.adoc
+++ b/downstream/modules/platform/con-edge-manager-build-bootc.adoc
@@ -1,7 +1,5 @@
 :_mod-docs-content-type: CONCEPT
 
-:_mod-docs-content-type: <CONCEPT>
-
 [id="edge-manager-build-bootc"]
 
 = Building a _bootc_ operating system image for the {RedHatEdge}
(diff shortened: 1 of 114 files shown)


downstream/assemblies/platform/assembly-edge-manager-images.adoc
```diff
@@ -31,7 +31,7 @@ include::platform/con-edge-manager-drop-dir.adoc[leveloffset=+2]
 
 include::platform/con-edge-manager-os-img-script.adoc[leveloffset=+2]
 
-include::platform/con-edge-manager-build-bootc.adoc[leveloffset=+1]
```
michellemacrh, line 34: This doesn't seem to be mod docs content type related - is that right?

downstream/modules/platform/con-edge-manager-build-bootc.adoc
```diff
@@ -1,3 +1,5 @@
+:_mod-docs-content-type: CONCEPT
+
 :_mod-docs-content-type: <CONCEPT>
```
michellemacrh, line 3: This line may want to be removed so it doesn't cause a conflict

downstream/assemblies/platform/assembly-edge-manager-images.adoc
g-murray: Yeah I think this was a slip up when I was going back and removing the incoorect usage of < > this was just collateral damage :(  Fixed now though!

PR #3829, opened 2025-07-14
2.5 Update restore proc for containerized installation (#3828)

michellemacrh:
Backports #3828 from main to 2.5 

Containerized restores to another environment fails to find the restore files

https://issues.redhat.com/browse/AAP-45444


g-murray (approved):
LGTM

PR #3830, opened 2025-07-14
Updating migration section as per AAP-49300

g-murray:
[AAP-49300](https://issues.redhat.com/browse/AAP-49300)
Current AAP 2.5 RPM -> OpenShift Procedure does not work

Updating migration section as per AAP-49300


michellemacrh:
LGTM!

PR #3854, opened 2025-07-17
DITA migration changes: CUG Ch 23

ianf77:
New module. New graphics

Using automation execution UI and modular compliance chapter 23

https://issues.redhat.com/browse/AAP-45157


downstream/modules/platform/proc-get-started-credential-types.adoc
```diff
@@ -1,23 +1,25 @@
 :_mod-docs-content-type: REFERENCE
```
g-murray, line 1: Update REFERENCE to PROCEDURE

downstream/modules/platform/proc-get-started-credential-types.adoc
```diff
@@ -1,23 +1,25 @@
...
-[id="ref-get-started-credential-types"]
+[id="proc-get-started-credential-types"]
 
 = Getting started with credential types
 //[ddacosta] Consider rewriting this as a procedure.
 
 .Procedure
 . From the navigation panel, select {MenuAECredentials}.
++
 If no custom credential types have been created, the *Credential Types* page prompts you to add one.
```
g-murray, line 11: This might flag in VALE as it's not a step. "Content other than a single list cannot be mapped to DITA tasks."

Not 100% sure because you're using "+", you might get away with it

g-murray (approved):
One potential AEM migration issue, besides that LGTM

PR #3872, opened 2025-07-21
AAP-49823: Replace text with attributes and run vale dita check

jonquilwilliams:
For the Getting Started with Terraform guide, replaced some text with attributes and used the Vale DITA extension to check DITA compatibility.

https://issues.redhat.com/browse/AAP-49823


g-murray (approved):
LGTM

PR #3894, opened 2025-07-23
Convert link fragments to full links in preparation for DITA conversion

jhradilek:
This resolves the use case when attributes are treated as a full, reusable links and defined as such in attributes/attributes.adoc:

```asciidoc
:LinkBuilder: {URLBuilder}[{TitleBuilder}]
```

These were previously referenced as follows, which lead to invalid markup if the attribute couldn't be resolved:

```asciidoc
link:{LinkBuilder}
```

This pull request does not address the situation when links are composed from multiple fragments as part of the text:

```asciidoc
link:{BaseURL}/{PlatformVers}/book_title[Book Title]
```


ariordan-redhat (approved):
LGTM

PR #3945, opened 2025-07-31
Adjusting wording for 2.6 version

g-murray:
[AAP-48233](https://issues.redhat.com/browse/AAP-48233)
[2.6] OCP - Docs Support for PostgreSQL 16 & 17

Minor tweak after a chat with an SME

### DIFF
--- a/downstream/modules/platform/proc-operator-external-db-controller.adoc
+++ b/downstream/modules/platform/proc-operator-external-db-controller.adoc
@@ -24,9 +24,9 @@ The external database must be a PostgreSQL database that is the version supporte
 
 [NOTE]
 ====
-{PlatformNameShort} {PlatformVers} supports {PostgresVers} for its managed databases and PostgreSQL 16 and 17 for external databases.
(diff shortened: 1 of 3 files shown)


downstream/modules/platform/proc-operator-external-db-controller.adoc
```diff
@@ -24,9 +24,9 @@ The external database must be a PostgreSQL database that is the version supporte
 
 [NOTE]
 ====
-{PlatformNameShort} {PlatformVers} supports {PostgresVers} for its managed databases and PostgreSQL 16 and 17 for external databases.
+{PlatformNameShort} {PlatformVers} supports {PostgresVers} for its managed databases and additionally supports PostgreSQL 15, 16, and 17 for external databases.
 
-If you choose to use an externally managed database with version 16 or 17 you must also rely on external backup and restore processes.
+If you choose to use an externally managed database you must also rely on external backup and restore processes.
```
michellemacrh, line 29: I think the original version was correct, at least per my understanding: "If you choose to use an externally managed database with version 16 or 17 you must also rely on external backup and restore processes."

- If you use an external DB with PG 15, backup and restore works
- If you use an external DB with PG 16 or 17, you need to bring your own backup/restore solution

PR #4006, opened 2025-08-06
2.6 AAP-44556 Add registry check in devtools doc (#4000)

ariordan-redhat:
AAP-44556 Add registry check in devtools doc
Per the Jira issue, and after consultation with the Devtools team, removed references to Docker, because the rest of the doc refers to Podman only.


g-murray (approved):
LGTM

PR #4025, opened 2025-08-11
AAP-51375 Migration: Change devtools xrefs to links

ariordan-redhat:
AAP-51375 Migration: Change devtools xrefs to links


hherbly (approved):
LGTM

PR #4027, opened 2025-08-11
2.6 AAP-51375 Migration: Change devtools xrefs to links (#4025)

ariordan-redhat:
AAP-51375 Migration: Change devtools xrefs to links (#4025)


rogrange:
LGTM

PR #4079, opened 2025-08-15
Add empty line after conditional titles (#4077)

ianf77:
Correct error with conditionals for DITA migration

https://issues.redhat.com/browse/AAP-51625


EMcWhinn (approved):
LGTM

PR #4094, opened 2025-08-20
New Assembly and modules for OCP (AAP-51782)

g-murray:
[AAP-51782](https://issues.redhat.com/browse/AAP-51782)
Troubleshooting Documentation for AAP on OCP

### DIFF
--- a/downstream/modules/platform/proc-operator-config-log-verbosity.adoc
+++ b/downstream/modules/platform/proc-operator-config-log-verbosity.adoc
@@ -1,0 +1,34 @@
+:_mod-docs-content-type: PROCEDURE
+
+[id="proc-operator-config-log-verbosity_{context}"]
+
+= Configuring log verbosity
+
+You can enable task output for debugging on any custom resources (CRs) by setting `no_log` to `false` in the `spec` section of the CR.
+The logs then show output for any failed tasks that originally had `no_log` set to `true`.
+The following procedure uses {ControllerName} as an example, but every CR listed in the Core Ansible Automation Platform Resources section supports `no_log`. 
+
+.Procedure
+
+. Edit the {ControllerName} CR and set the `no_log` field to `false` in the spec.
++
+----
+apiVersion: automationcontroller.ansible.com/v1beta1
+kind: AutomationController
+metadata:
+  name: controller-demo
+spec:
+  no_log: false
+----
++
+[NOTE]
+====
+This might expose sensitive data in the logs. On production clusters, this value must generally be set to `true` unless you are actively debugging an issue.
+====
++
+. To increase the Ansible Playbook verbosity from the operator, set the verbosity level using an annotation:
++
+----
+annotations:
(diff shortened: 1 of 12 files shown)


downstream/modules/platform/proc-operator-config-log-verbosity.adoc
```diff
@@ -0,0 +1,34 @@
+:_mod-docs-content-type: PROCEDURE
+
+[id="proc-operator-config-log-verbosity_{context}"]
+
+= Configuring log verbosity
+
+You can enable task output for debugging on any custom resources (CRs) by setting `no_log` to `false` in the `spec` section of the CR.
+The logs then show output for any failed tasks that originally had `no_log` set to `true`.
+The following procedure uses {ControllerName} as an example, but every CR listed in the Core Ansible Automation Platform Resources section supports `no_log`.
```
michellemacrh, line 9: Could make use of an attribute here for "Ansible Automation Platform"
{PlatformNameShort}

downstream/modules/platform/proc-viewing-operator-logs.adoc
```diff
@@ -0,0 +1,28 @@
...
+
+.Procedure
+
+. To find the pod name, run:
++
+----
+oc get pods | grep operator
+----
++
+. View the logs for the pod, run:
```
michellemacrh, line 17: This is a really tiny nit so completely optional but I wonder if this reads better as "To view the logs for the pod, run:" - The first step in the proc starts with 2 so this keeps them sort of in parallel.

downstream/modules/platform/ref-operator-standard-k8s-resources.adoc
```diff
@@ -0,0 +1,37 @@
+:_mod-docs-content-type: REFERENCE
+
+[id="ref-operator-standard-k8s-resources_{context}"]
+
+= Standard Kubernetes  resources
```
michellemacrh, line 5: I don't think this will effect anything but looks like there's an extra space here: "Kubernetes  resources"

downstream/modules/platform/proc-viewing-operator-logs.adoc
```diff
@@ -0,0 +1,28 @@
...
++
+----
+oc logs <operator-pod-name> -f
+----
++
+* Alternatively, to view the logs without first getting the pod name, run:
```
michellemacrh, line 23: This is an unordered list item but the other two are ordered list items - not sure if this is intentional but just calling it out.

michellemacrh (commented):
A few small notes otherwise LGTM!

downstream/modules/platform/proc-operator-config-log-verbosity.adoc
g-murray: Opps I forgot to comment that I plan to add a link to this section after it's created. So I will need to back and tweak afterwards, this was a brilliant reminder as I had forgotten :P

michellemacrh:
LGTM

PR #4127, opened 2025-08-26
Editing controller files for DITA

EMcWhinn:
Configuring automation execution modular compliance chapters 19/20

https://issues.redhat.com/browse/AAP-46738


downstream/modules/platform/proc-controller-ec2-vpc-instances.adoc
```diff
@@ -18,7 +19,9 @@ vpc_destination_variable: private_ip_address
 +
 . Click btn:[Save] and trigger an update of the group.
 
-Once this is done you can see your VPC instances.
+.Next steps
+
+Once you complete this steps, you can see your VPC instances.
```
g-murray, line 24: These* steps. Also would this be a verification step rather than Next steps?

downstream/modules/platform/proc-controller-config-jump-hosts-inventory.adoc
```diff
@@ -0,0 +1,22 @@
+:_mod-docs-content-type: PROCEDURE
+
+[id="controller-config-jump-hosts-inventory"]
+
+= Configuring jump hosts using Ansible Inventory variables
+
+You can also add a jump host to your {ControllerName} instance through Inventory variables.
```
g-murray, line 7: From a minimalism pov I think you could get away without using "also"

g-murray (commented):
One or two suggestions otherwise LGTM

g-murray (approved):
LGTM

PR #4143, opened 2025-08-27
Updating team mapping section (#4142)

EMcWhinn:
[Docs] Authentication mapping for teams does not work if "join_condition": "and" with attributes

https://issues.redhat.com/browse/AAP-51465

Affects `titles/central-auth`


g-murray (approved):
LGTM

PR #4178, opened 2025-09-02
Add procedure to use custom inventory plugin

ianf77:
[Docs] Adding Guide for Using Inventory Sync Plugins from Multiple Collections

https://issues.redhat.com/browse/AAP-43019


g-murray (approved):
LGTM

PR #4315, opened 2025-09-17
Update Support matrix for upgrade scenarios

michellemacrh:
Affected title: `planning-your-upgrade`

- The usage of inline links was confirmed with PM.
- Links included to different AAP product versions is intentional.

Update the Support matrix for upgrade scenarios "Process" column to include inline links to the relevant documentation

https://issues.redhat.com/browse/AAP-53516

### DIFF
--- a/downstream/modules/platform/ref-upgrade-scenarios-container.adoc
+++ b/downstream/modules/platform/ref-upgrade-scenarios-container.adoc
@@ -16,7 +16,7 @@ Use the following reference tables to find the supported upgrade paths for conta
 | Container-based {PlatformNameShort} 2.5 on RHEL 9 
 | Container-based {PlatformNameShort} 2.6 on RHEL 9 
 a| 
-. link:{URLContainerizedInstall}/aap-containerized-installation#updating-containerized-ansible-automation-platform[Upgrade your container deployment] from 2.5 to 2.6.
+* link:{URLContainerizedInstall}/aap-containerized-installation#updating-containerized-ansible-automation-platform[Upgrade your container deployment] from 2.5 to 2.6.
 
 | Container-based {PlatformNameShort} 2.5 on RHEL 9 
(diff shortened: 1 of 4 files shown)


downstream/modules/platform/ref-upgrade-scenarios-container.adoc
```diff
@@ -16,19 +16,19 @@ Use the following reference tables to find the supported upgrade paths for conta
 | Container-based {PlatformNameShort} 2.5 on RHEL 9 
 | Container-based {PlatformNameShort} 2.6 on RHEL 9 
 a| 
-. Upgrade your container deployment from 2.5 to 2.6.
+. link:{URLContainerizedInstall}/aap-containerized-installation#updating-containerized-ansible-automation-platform[Upgrade your container deployment] from 2.5 to 2.6.
```
EMcWhinn, line 19: I'm just wondering in the `Process` column with one link do you want a bullet point rather than a numbered list? In the build this shows as "1.Upgrade your container deployment from 2.5 to 2.6."

EMcWhinn (commented):
One question, otherwise LGTM.

downstream/modules/platform/ref-upgrade-scenarios-container.adoc
michellemacrh: hhmm let me double check - I think they should all be numbered lists 🤔

michellemacrh: Oh I get what you mean - good point 😮‍💨 I wonder if it's only one step does it need a bullet point/number at all?

EMcWhinn: Ya that's what I meant just the single ones, I'd say even removing the list ya would work.

PR #4239, opened 2025-09-09
Correct statements in API guide

ianf77:
Bug fix: create token procedure needs to include Gateway

https://issues.redhat.com/browse/AAP-51975

PR #4321, opened 2025-09-17
Correct statements in API guide (#4239)

ianf77:
Bug fix: create token procedure needs to include Gateway

https://issues.redhat.com/browse/AAP-51975

### DIFF
--- a/downstream/modules/platform/con-controller-api-oauth2-token.adoc
+++ b/downstream/modules/platform/con-controller-api-oauth2-token.adoc
@@ -34,10 +34,7 @@ Token authentication is best used for any programmatic use of the {ControllerNam
 
 .Curl example
 
-[literal, options="nowrap" subs="+attributes"]
-----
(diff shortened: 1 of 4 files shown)


downstream/modules/platform/con-controller-api-oauth2-token.adoc
```diff
@@ -34,10 +34,7 @@ Token authentication is best used for any programmatic use of the {ControllerNam
 
 .Curl example
 
-[literal, options="nowrap" subs="+attributes"]
-----
-curl -u user:password -k -X POST https://<controller-host>/api/v2/tokens/
-----
+'curl -u user:password -k -X POST https://<platform-host>/api/gateway/v1/tokens/`
```
EMcWhinn, line 37: I just noticed that the beginning of this has ' instead of ` so it won't render as code. I'll merge the backports and fix this in my open PR.

EMcWhinn (approved):
LGTM

PR #4339, opened 2025-09-18
Removing legacy auth step from Access Management

EMcWhinn:
[DOCS] 2.6 - Remove legacy auth step from auth type procedures

https://issues.redhat.com/browse/AAP-53902

Affects `titles/central-auth`

PR #4346, opened 2025-09-19
Removing legacy auth step from Access Management (#4339)

EMcWhinn:
* Removing legacy auth step from access management

[DOCS] 2.6 - Remove legacy auth step from auth type procedures

https://issues.redhat.com/browse/AAP-53902

Affects `titles/central-auth`

* Adding PAT migration module

* Adding attribute

* Peer review edits


sayjadha (commented):
LGTM.

PR #4355, opened 2025-09-19
AAP-54033: Fix broken links (#4352)

jonquilwilliams:
* Fix Vault links (I needed to add vault-product# in front of the topic ID, so the format is:
`link:{URLHashiGuide}/vault-product#<topidId>[<topicName>]`)

* Replaced 2.5 with version attribute {PlatformVers}

Backport for #4352 

The links probably won't work until 2.6 is published. But the paths are correct and the {PlatformVers} does point correctly in the bottom left corner when it tries to fetch that page.

It also works correctly in the 2.5 backport since that content is published.


rogrange:
LGTM

PR #4426, opened 2025-09-24
Backport 2.6 for 4416 (Updates to Planning Guide Overview)

jonquilwilliams:
https://issues.redhat.com/browse/AAP-54344

Backport for #4416


downstream/modules/platform/con-aap-upgrade-overview.adoc
rogrange, line 1: The only change I would make is to replace the 2.6 references with {PlatformVers}. Other than that, it looks good.

PR #4428, opened 2025-09-24
AAP-46607-rpmupgrade: Added info. about EDA 2.4 upgrade

sayjadha:
This PR addresses JIRA https://issues.redhat.com/browse/AAP-46607. 

Changes made:

- Added procedure about removing EDA 2.4 DB before upgrading to AAP 2.6. 
- Updated the necessary notes to recommend the procedure.

### DIFF
--- a/downstream/modules/platform/con-aap-upgrades.adoc
+++ b/downstream/modules/platform/con-aap-upgrades.adoc
@@ -8,7 +8,7 @@ Currently, it is possible to perform {PlatformNameShort} upgrades using one of t
 
 [IMPORTANT]
 ====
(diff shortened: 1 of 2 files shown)


downstream/modules/platform/con-aap-upgrades.adoc
rogrange, line 1: "2.5 to 2.6 is supported; however, upgrade from {EDAName} 2.4 to 2.6"

"to 2.6 and you have"

"your platform to 2.6."

Same question.

downstream/modules/platform/proc-upgrade-controller-hub-eda-unified-services.adoc
rogrange, line 1: "data into the 2.6 inventory."
"unified UI ({Gateway}) for 2.6

same question.

PR #4508, opened 2025-10-01
Post release update for planning your upgrade guide

g-murray:
Just making a change now the section exists.


downstream/assemblies/platform/assembly-upgrade-faqs.adoc
rogrange, line 1: suggestion, replace 
"what's" with "what is" and we are good.

PR #4566, opened 2025-10-03
AAP 55092: Fix Xrefs in GS with Hasi/AAP guide

jonquilwilliams:
https://issues.redhat.com/browse/AAP-55092

Xrefs were broken due to a funky TOC change. They work now in the preview.


hherbly (approved):
LGTM

PR #4632, opened 2025-10-08
AAP-55501 Update Portal upgrade process

ariordan-redhat:
AAP-55501 Update the Portal upgrade process.

Affects `titles/self-service-install`


g-murray:
Handled in https://github.com/ansible/aap-docs/pull/5178/

PR #5889, opened 2026-04-06
AAP-69732: removes references to API tokens in hub docs

hherbly:
This work is being tracked in [AAP-69732](https://redhat.atlassian.net/browse/AAP-69732)


hherbly:
AAP 2.6 docs were migrated away from Github and are being published from AEM. This content removal occurred on that platform.

PR #5915, opened 2026-04-09
AAP-70258 Add database storage requirements for OCP deployment models

g-murray:
Add storage sizing guidance that was missing from the initial PR:
- Operator-deployed DB (OCP-A): 100 GB hard limit with storage condition
- External DB (OCP-B): 200 GB minimum on 3000 IOPS storage


KylaF8 (approved):
LGTM!

PR #5921, opened 2026-04-10
Fixing the title to match with the other titles. (#5893)

KylaF8:
AAP-69877: https://redhat.atlassian.net/browse/AAP-69877

* Fixing the title to match with the other titles.

* Changed it back.

* Removing the attribute to try and fix the heading.


g-murray (approved):
LGTM

PR #5923, opened 2026-04-13
Update executtion precedence appendix

ianf77:
Update documentation to make it clear that "Default Execution Environment" may not be used in certain situations

https://redhat.atlassian.net/browse/AAP-66680


jself-sudoku:
lgtm

PR #5927, opened 2026-04-15
Highlight upgrade and migration changes

ianf77:
Planning your upgrade and migration content highlights that customers need to be on either Containerized or OCP deployment before considering upgrade to 2.7

https://redhat.atlassian.net/browse/AAP-66569


g-murray (approved):
LGTM

rooftopcellist (approved):

PR #5936, opened 2026-04-21
Adding deprecated note to RHEM doc

EMcWhinn:
Add deprecation notice to AAP Edge Manager guides

https://redhat.atlassian.net/browse/AAP-72427


KylaF8 (approved):
LGTM!

PR #5946, opened 2026-04-23
Corrections to mcp/execution environments

ianf77:
Testathon changes to MCP servers and execution environments.

https://redhat.atlassian.net/browse/AAP-72785

### DIFF
--- a/downstream/assemblies/builder/assembly-MCP-servers-and-EE.adoc
+++ b/downstream/assemblies/builder/assembly-MCP-servers-and-EE.adoc
@@ -9,14 +9,14 @@ A _Model Context Protocol_ (MCP) Server is an external service that acts as a ga
 
 An MCP eliminates the need for custom integrations by providing a universal "language" for AI models to access and interact with real-world systems, data, and tools.
 
-The MCP operates using a client-server architecture:
+The MCP operates using a client/server architecture:
 
 [cols="a,a,a", options="header"]
 |===
 | Component | Role | Description
-| MCP Host | The application environment | This is where the user interacts with the AI, such as an AI-powered IDE like VS Code or a chatbot interface.
+| MCP Host | The application environment | This is where the user interacts with the AI, such as an AI-powered IDE like Visual Studio Code or a chatbot interface.
 | MCP Client | Connection manager | A component within the Host that translates LLM requests into the MCP format.
-| MCP Server | Resource provider | An external service that connects to underlying systems like databases or APIs.
+| MCP Server | Resource provider | An external service that connects to underlying systems such as databases or APIs.
 |===
 
 By using this standardized protocol, any LLM that implements the MCP Client can use any MCP Server.
--- a/downstream/modules/builder/proc-build-ee-with-mcp-server.adoc
+++ b/downstream/modules/builder/proc-build-ee-with-mcp-server.adoc
@@ -96,6 +96,3 @@ A successful build produces output ending with:
 Successfully tagged localhost/my-cfn-mcp-ee:latest
 Complete! The build context can be found at: /path/to/ee-build/context
 ----
-
-
-
--- a/downstream/modules/builder/proc-configure-remote-mcp-server.adoc
+++ b/downstream/modules/builder/proc-configure-remote-mcp-server.adoc
@@ -7,7 +7,7 @@
 [role="_abstract"]
 If you need to bypass local installation and dynamically choose your deployment mode at build time, configure your role to reference a remote MCP server over HTTP.
 
-.Add a remote server to your role
+*Add a remote server to your role*
 To define a remote MCP server, set the type to http in your registry metadata:
 
 ----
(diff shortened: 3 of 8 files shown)


fistilis:
Hey Ian, thanks for the PR. I reviewed against the feedback from AAP-73188 and the testathon notes — a few items need attention before this can merge.

## Regressions introduced by this PR

1. **`pyoi` typo introduced**: In `ref-evaluate-mcp-servers.adoc`, the change goes from `lang: "pypi"` *to* `lang: "pyoi"` — this introduces the typo into the security best practices section rather than fixing it. The original `pyoi` typo in the installation method table (in `ref-mcp-builder-collection.adoc`) also still needs to be corrected to `pypi`.

2. **`mkdir` command is still wrong**: The fix uses square brackets `[defaults,tasks,meta]` instead of curly braces `{defaults,tasks,meta}`, and removes `roles/cfn_mcp/` from the path. The correct command should be:
   ```
   mkdir -p collections/ansible_collections/myorg/mcp_cfn/roles/cfn_mcp/{defaults,tasks,meta}
   ```

3. **Version string `1.0.1.9` should be `1.0.19`**: All version updates use a four-segment version (`1.0.1.9`) instead of the correct three-segment semver `1.0.19`. Affects galaxy.yml, defaults/main.yml, dependency field, and security best practices.

4. **Final tarball reference still says `1.0.0`**: End of `proc-create-custom-mcp-server-role.adoc` still has `myorg-mcp_cfn-1.0.0.tar.gz` — should be `1.0.19` to match.

5. **Mismatched quotes on manifest path**: The fix for the concatenated path uses a backtick to open and a single quote to close: `` `/opt/mcp/mcpservers.json') ``. Should use matching backticks.

6. **Double backtick still present**: `install_manager`` task` — the errant double backtick from the original is still there.

## Still outstanding from AAP-73188

- **"built-in" → "dev preview"**: Section 6.6 references "built-in Model Context Protocol (MCP)" — should say "dev preview" per product positioning.
- **Invalid `lang` type downstream failure**: Docs say "install manager skips installation" but `generate_manifest.yml` then fails with `'dict object' has no attribute 'path'` (rc=2). Either document this downstream failure or note it as a known limitation.
- **Directory structure shows files no command creates**: The structure diagram shows galaxy.yml, README.md, LICENSE, and main.yml files, but no documented command creates them. This is misleading for someone following step-by-step.
- **Role name inconsistency**: Role name switches from `cfn_mcp` to `mcp_cfn` at step 7, breaking copy-paste flow.
- **AsciiDoc rendering issues**: Section 6.7 has "ote" (part of "remote") greyed out; sections 6.8 and 6.11 have +/* characters that should render as bullets.
- **Remote mode wording contradiction**: The main text still says "skips installation" right before the new NOTE saying the package *is* still installed. Suggest rewording the main text itself rather than contradicting it with a note.

fistilis:
## Review of May 7 PDF build against prior feedback

Reviewed the latest PDF export against the comments left on this PR and AAP-73188. Here's what I found:

### Fixed
- **pyoi typo** — Now correctly shows `lang: "pypi"` in the installation method table and all code blocks.
- **Remote mode contradiction** — Section 10 no longer says "skips installation" before a NOTE contradicting it. Reworded to "bypass local installation" with consistent explanation.

### Partially Fixed
- **"built-in" vs "dev preview"** — Section 7 title correctly says "dev preview" and the support boundary section (p28) says "Dev Preview." However, the NOTE on p12 still reads "The **built-in** `ansible.mcp_builder.install_mcp` playbook…" and the troubleshooting intro (section 13) still says "custom or **built-in** Model Context Protocol (MCP) servers." These should say "dev preview" for consistency.
- **Invalid lang type failure** — The troubleshooting table now has an "Invalid language type" row, but it only lists the valid values (pypi, npm, go). It doesn't describe the actual failure behavior — does the build fail silently? Does ansible-playbook error out? What task fails? Users hitting this need to know what symptom to look for.

### Not Fixed
- **Version string `1.0.1.9`** — Still four-segment in every occurrence (p3, p8 galaxy.yml, p8 defaults/main.yml, p8 dependencies, p13 tarball). Should be three-segment semver (`1.0.19`). Additionally, the **tarball version mismatch** persists: step 7 produces `myorg-mcp_cfn-1.0.1.9.tar.gz` but section 6 step 1 copies `myorg-mcp_cfn-1.0.0.tar.gz`. A user following these steps sequentially would get a "file not found" error.
- **Directory structure vs. commands mismatch** — Step 1 (section 5) has 4 `touch` commands, but the directory tree shows 7 files. Missing from the commands: `playbooks/install_cfn_mcp.yml`, `roles/cfn_mcp/defaults/main.yml`, `roles/cfn_mcp/tasks/main.yml`. Also, `touch` won't create parent directories — these commands will fail on a clean system. Needs `mkdir -p` first, or use `mkdir -p collections/ansible_collections/myorg/mcp_cfn/roles/cfn_mcp/{defaults,tasks,meta}` followed by the touch commands.
- **Role name inconsistency** — Collection is `mcp_cfn`, role is `cfn_mcp`. Technically consistent but very confusing for users, especially in a tutorial. Consider aligning these (e.g., both `cfn_mcp`).

### Cannot Verify from PDF (need AsciiDoc source)
- **Mismatched quotes on manifest path** (backtick open, single-quote close)
- **Double backtick on `install_manager``**
- **AsciiDoc rendering artifacts** in sections that mapped to 6.7, 6.8, 6.11 (greyed text, `+`/`*` artifacts) — PDF renders clean but HTML output may still have issues

PR #5945, opened 2026-04-23
AAP-72447: Clarify single-org support in self-service portal docs

jklech-work:
## Summary

- Change plural "Organizations" to singular across all sync-related descriptions in the self-service portal guides
- Only one AAP Organization is currently supported; multi-org is on the roadmap (ANSTRAT-912)
- Fixes 7 occurrences across 5 files that implied multi-org support, which was misleading customers

## Test plan

- [x] `self-service-config` guide builds clean with ccutil
- [x] Peer review: verify singular form reads naturally in context
- [x] SME review: Craig Brandt confirmed scope

## Jira

[AAP-72447](https://issues.redhat.com/browse/AAP-72447)

PR #5949, opened 2026-04-24
(2.6 backport) AAP-72447: Clarify single-org support in self-service portal docs

jklech-work:
backports #5945


michellemacrh (approved):
LGTM

PR #5952, opened 2026-04-28
AAP-71980: Updated Ansible ALIA documentation for Azure OpenAI configuration

sayjadha:
This PR addresses JIRA https://redhat.atlassian.net/browse/AAP-71980.

Changes made:
- Added the parameter value requirement for Azure OpenAI configuration.

### DIFF
--- a/downstream/modules/platform/proc-configure-lightspeed-variables-container-install.adoc
+++ b/downstream/modules/platform/proc-configure-lightspeed-variables-container-install.adoc
@@ -41,7 +41,7 @@ lightspeed_chatbot_model_api_key: <set your own>
 lightspeed_chatbot_model_id: : <set your own>
 lightspeed_chatbot_default_provider: 'rhoai'
 lightspeed_chatbot_model_extra_settings: {} 
-# If you want to use {AzureOpenAI} as the LLM provider, specify the lightspeed_chatbot_model_extra_settings value as '{"api_type": ""}'. 
+# If you want to use Microsoft Azure OpenAI as the LLM provider, specify the lightspeed_chatbot_model_extra_settings value as '{"api_type": ""}', and the lightspeed_chatbot_model_url value to 'https://your_inference_api/openai/v1'. 
 
 # This section is to configure Ansible Lightspeed intelligent assistant with MCP server integration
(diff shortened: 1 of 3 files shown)


downstream/modules/platform/proc-configure-lightspeed-variables-container-install.adoc
```diff
@@ -41,7 +41,7 @@ lightspeed_chatbot_model_api_key: <set your own>
 lightspeed_chatbot_model_id: : <set your own>
 lightspeed_chatbot_default_provider: 'rhoai'
 lightspeed_chatbot_model_extra_settings: {} 
-# If you want to use {AzureOpenAI} as the LLM provider, specify the lightspeed_chatbot_model_extra_settings value as '{"api_type": ""}'. 
+# If you want to use Microsoft Azure OpenAI as the LLM provider, specify the lightspeed_chatbot_model_extra_settings value as '{"api_type": ""}', and the lightspeed_chatbot_model_url value to 'https://your_inference_api/openai/v1'.
```
g-murray, line 44: Two comments here, is there a reason you removed the attribute here and not in the other two files. Second, I think "lightspeed_chatbot_model_url" should also have backticks right? (not just the URL)

g-murray (commented):
Just two questions for you on this

downstream/modules/platform/proc-configure-lightspeed-variables-container-install.adoc
sayjadha: @g-murray answers below:
#1: I removed the  {AzureOpenAI} from the code block snippet, but retained it in other places. The attribute didn't get rendered correctly as it was in a code block. Here's the published output: https://docs.redhat.com/en/documentation/red_hat_ansible_automation_platform/2.6/html-single/containerized_installation/index#proc-configure-lightspeed-variables_deploying-lightspeed-containerized-install

#2: The parameters are not in backticks in the code snippet, the example values are. See the same doc above for example.

g-murray: Oh sorry I didn't spot this was a code block

g-murray (approved):
LGTM

PR #5963, opened 2026-04-29
Comment out EE Builder docs - feature not yet public

jklech-work:
## Summary

- EE Builder feature is not public or ready for customers per Craig Brandt
- Commented out 3 include directives across 2 guides to remove EE Builder content from published docs
- Content remains on main for future re-enablement when the feature is ready

### Changes

**self-service-config guide:**
- Commented out repository synchronization chapter

**self-service-using guide:**
- Commented out creating EE definitions chapter
- Commented out discovering EE content chapter

## Verification
- Both `self-service-config` and `self-service-using` guides build successfully with these changes

Companion PR for 2.6: #5962


ianf77 (approved):
LGTM

PR #5964, opened 2026-04-29
Comment out EE Builder disconnected import from content setup chapter

jklech-work:
## Summary

- Follow-up to #5962: removes remaining EE Builder reference from the content setup chapter
- The `proc-self-service-import-templates-disconnected.adoc` module ("Import EE Builder templates in disconnected environments") was still included in the content setup assembly
- Commented out the include on the 2.6 branch — main stays untouched for future re-enablement

## Verification
- `self-service-config` guide builds successfully


g-murray (approved):
LGTM

PR #5966, opened 2026-04-29
Edits to the new assembly,reference, and procedure.

KylaF8:
AAP-39131 - https://redhat.atlassian.net/browse/AAP-39131


hherbly (approved):
LGTM

PR #5973, opened 2026-04-30
Comment out content management chapter in self-service config guide

jklech-work:
## Summary
- Comments out Chapter 7 "Set up content management in self-service automation portal" from the configuring guide per Craig's guidance — this content should not be published on the customer portal

## Test plan
- [x] self-service-config guide builds successfully


jklech-work:
Wrong target — this should only be on release branches (2.5/2.6), not main.

PR #5974, opened 2026-04-30
Comment out content management chapter in self-service config guide

jklech-work:
## Summary
- Comments out Chapter 7 "Set up content management in self-service automation portal" from the configuring guide per Craig Brandt's guidance — this content should not be published on the customer portal yet

## Test plan
- [x] Guide builds successfully


EMcWhinn (approved):
LGTM

PR #5995, opened 2026-05-27
AAP-70906: adds deprecation notice to ALIA 2.5 docs

hherbly:
This work is being tracked in [AAP-70906](https://redhat.atlassian.net/browse/AAP-70906) and applies to the 2.5 docs only.


jonquilwilliams:
lgtm

PR #5998, opened 2026-06-01
Fixed build error

ianf77:
Replaced placemarker with xref

Fix build error in 2.5 Using automation analytics

https://redhat.atlassian.net/browse/AAP-77236


hherbly (approved):
LGTM

PR #6041, opened 2026-08-04
Updated private/public azure images

ianf77:
Update Azure Network Architecture Diagram — Rename Subnets, Fix IP Consistency, Add Private vs Public Clarity

https://redhat.atlassian.net/browse/AAPRFE-2968


michellemacrh (approved):
LGTM

PR #6044, opened 2026-08-07
Update AAP with new vm series

ianf77:
Update AAP on Azure docs with new vm series

https://redhat.atlassian.net/browse/AAP-84330


jself-sudoku:
lgtm

PR #6057, opened 2026-08-24
AAP-88332 Add responsibility matrix to AAP on Azure docs

g-murray:
## Summary
- Adds a new reference topic (`ref-azure-responsibilities.adoc`) containing the responsibility matrix from KB article 6966926, covering support and change management responsibilities across Customer, Red Hat, and Microsoft
- Includes sections on security/compliance posture and customer responsibilities
- Includes the new topic in the existing `assembly-azure-support.adoc` assembly

## Test plan
- [ ] Verify the topic renders correctly in a local preview build
- [ ] Confirm the responsibility table matches the source KB article (https://access.redhat.com/articles/6966926)
- [ ] Check the `con-azure-lifecycle` cross-reference resolves correctly
- [ ] Verify product name attributes render as expected

Resolves: [AAP-88332](https://redhat.atlassian.net/browse/AAP-88332)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #6060, opened 2026-08-24
Fix broken xref to con-azure-lifecycle in responsibilities matrix

g-murray:
## Summary

- Fix broken cross-reference to `con-azure-lifecycle` in `ref-azure-responsibilities.adoc` by adding the missing `_azure-intro` context suffix

The xref was introduced in #6057 (AAP-88332) without the context suffix that the target topic's ID requires.

## Test plan

- [ ] Verify the build no longer reports `Unknown ID or title "con-azure-lifecycle"`
- [ ] Verify the xref resolves correctly in the rendered output

🤖 Generated with [Claude Code](https://claude.com/claude-code)


michellemacrh (approved):
LGTM

PR #6063, opened 2026-09-02
Add deprovsion info for containerized

ianf77:
Exact steps to remove/deprovision nodes in containerized AAP installation is missing from the docs.

https://redhat.atlassian.net/browse/AAP-83204

### DIFF
--- a/downstream/modules/platform/proc-deprovision-groups-in-containerized.adoc
+++ b/downstream/modules/platform/proc-deprovision-groups-in-containerized.adoc
@@ -5,20 +5,17 @@
 = Deprovision instance groups in containerized deployments
 
 [role="_abstract"]
-Deprovisioning nodes removes them from the mesh. For instance groups, unregister the group from automation controller's cluster registry by using the `awx-manage unregister_queue` command. This removes the group definition but does not deprovision its member nodes.
+Unregister instance groups by using the `awx-manage` command to cleanly remove group definitions from a containerized {PlatformNameShort} deployment without impacting member nodes.
 
 .Prerequisites
 
-* You have root or sudo access to the host running the automation controller containers.
+* You have root or sudo access to the host running the {ControllerName} containers.
 * You know the name of the instance group to remove.
 * No active jobs are running on nodes that are members of the instance group.
 
 .Procedure
 
 . Ensure that no jobs are currently running on instances in the group. In the automation controller UI, navigate to {MenuAEJobs} and confirm that no jobs with a status of `Running` or `Pending` are assigned to instances in the group.
-
-. (Optional) If the instance group contains nodes that should also be deprovisioned, deprovision each node using the procedure in
-xref:proc-deprovision-nodes-in-containerized[Deprovision nodes in a containerized deployment]. Deprovisioning a node automatically removes its membership from any instance groups, so no specific order of operations is required.
 
 . On the primary automation controller node, run the unregister command inside the automation controller task container. Replace `<group_name>` with the name of the instance group:
 +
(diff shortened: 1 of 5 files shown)


downstream/modules/platform/proc-deprovision-groups-in-containerized.adoc
```diff
@@ -0,0 +1,35 @@
+:_mod-docs-content-type: PROCEDURE
+
+[id="proc-deprovision-groups-in-containerized"]
+
+= Deprovision instance groups in containerized deployments
+
+[role="_abstract"]
+Deprovisioning nodes removes them from the mesh. For instance groups, unregister the group from automation controller's cluster registry by using the `awx-manage unregister_queue` command. This removes the group definition but does not deprovision its member nodes.
```
g-murray: The short description opens with "Deprovisioning nodes removes them from the mesh" but this topic is about *instance groups*, not nodes. The lead-in sentence is misleading.

Suggested rewrite:

> You can remove an instance group from the automation controller cluster registry by using the `awx-manage unregister_queue` command. This removes the group definition but does not deprovision its member nodes.

downstream/modules/platform/proc-deprovision-groups-in-containerized.adoc
```diff
@@ -0,0 +1,35 @@
...
+xref:proc-deprovision-nodes-in-containerized[Deprovision nodes in a containerized deployment]. Deprovisioning a node automatically removes its membership from any instance groups, so no specific order of operations is required.
+
+. On the primary automation controller node, run the unregister command inside the automation controller task container. Replace `<group_name>` with the name of the instance group:
++
+`podman exec -it automation-controller-task awx-manage unregister_queue --queuename=<group_name>`
++
+For example:
++
+`podman exec -it automation-controller-task awx-manage unregister_queue --queuename=remote_execution_group`
+
```
g-murray, line 27: `{MenuInfrastructureInstances}` points to the **Instances** page, not Instance Groups. This should be `{MenuInfrastructureInstanceGroups}` (or the equivalent attribute) to match the step's intent of verifying the *group* was removed.

downstream/modules/platform/proc-deprovision-groups-in-containerized.adoc
```diff
@@ -0,0 +1,35 @@
...
+. (Optional) If the instance group contains nodes that should also be deprovisioned, deprovision each node using the procedure in
+xref:proc-deprovision-nodes-in-containerized[Deprovision nodes in a containerized deployment]. Deprovisioning a node automatically removes its membership from any instance groups, so no specific order of operations is required.
+
+. On the primary automation controller node, run the unregister command inside the automation controller task container. Replace `<group_name>` with the name of the instance group:
++
+`podman exec -it automation-controller-task awx-manage unregister_queue --queuename=<group_name>`
++
+For example:
++
```
g-murray, line 25: This step says "or by using the API" but the sub-step only shows UI navigation — no API endpoint or command is provided. Either add the API verification method (e.g., `curl` to `/api/v2/instance_groups/`) or remove the "or by using the API" clause to avoid a dead reference.

downstream/modules/platform/proc-deprovision-groups-in-containerized.adoc
```diff
@@ -0,0 +1,35 @@
...
+* You know the name of the instance group to remove.
+* No active jobs are running on nodes that are members of the instance group.
+
+.Procedure
+
+. Ensure that no jobs are currently running on instances in the group. In the automation controller UI, navigate to {MenuAEJobs} and confirm that no jobs with a status of `Running` or `Pending` are assigned to instances in the group.
+
+. (Optional) If the instance group contains nodes that should also be deprovisioned, deprovision each node using the procedure in
+xref:proc-deprovision-nodes-in-containerized[Deprovision nodes in a containerized deployment]. Deprovisioning a node automatically removes its membership from any instance groups, so no specific order of operations is required.
+
```
g-murray: The note "no specific order of operations is required" is confusing here because the numbered steps imply a sequence (this is step 2, before the unregister command in step 3). If order truly doesn't matter, consider reordering so this optional node-deprovisioning step comes *after* the group unregister command, or reword to make it clear this is an independent action that can be done before or after removing the group.

g-murray (approved):
LGTM

